PR #2, opened 2023-06-18
Add CI with Github workflows

Laupetin:


diamante0018:
gcc is way more strict than msvc, I discovered that too some time ago. they basically force you to include each header. it seems the header containing numeric limits is missing somewhere

Laupetin:
> gcc is way more strict than msvc, I discovered that too some time ago. they basically force you to include each header. it seems the header containing numeric limits is missing somewhere

thanks,
yeah i had a pipeline with a linux build for some time now. i don't it's because it's more strict but because the headers of the runtime include other headers so in msvc these headers are included while with g++ they are not.
i guess github uses a different version so a lot of errors pop up that i previously didnt notice.

fixing all of them locally now because github actions only tells me about only one each time it runs and there might be a lot.

diamante0018:
good luck finding all of them. 🛩️

Issue #27, opened 2023-10-21
IWD project type


JezuzLizard:
In CoDs older than T6 IWDs were used to store assets for raw loading such as rawfiles, scripts, images, streamed sounds etc. An IWD is just a normal zip file compressed using zlib, just with .iwd extension. While not strictly necessary to add since it is trivial to just zip up the files and rename the file extension to iwd, the canonical mod tools do have an iwd list for the linker to know what files go into iwds and then the linker will create the iwd with the assets.

PR #66, opened 2023-12-23
Iwd project type

JezuzLizard:
Resolves issue #27 by adding the ability to create an iwd archive automatically using the linker's project system.


src/ObjWriting/ObjContainer/IWD/IWDWriter.cpp
```diff
@@ -0,0 +1,106 @@
...
+            return false;
+
+        AddFileToArchive(fileName, fileData, fileSize);
+
+        return true;
+    }
+
+    bool Write() override
+    {
+        const auto result = std::all_of(m_files.begin(),
```
diamante0018, line 83: Could use cbeing and cend in this case I think. No functional change

diamante0018 (changes requested):

src/Linker/Linker.cpp
```diff
@@ -45,15 +46,11 @@ enum class ProjectType
...
+    IWD,
     MAX
 };
 
-constexpr const char* PROJECT_TYPE_NAMES[static_cast<unsigned>(ProjectType::MAX)]{
-    "none",
-    "fastfile",
-    "ipak",
-};
+constexpr const char* PROJECT_TYPE_NAMES[static_cast<unsigned>(ProjectType::MAX)]{"none", "fastfile", "ipak", "iwd"};
```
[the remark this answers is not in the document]
JezuzLizard: Addressed.

src/Linker/Linker.cpp
```diff
@@ -462,6 +459,32 @@ class LinkerImpl final : public Linker
...
+
+        const auto iwdWriter = IWDWriter::Create(iwdFilePath.string(), &assetSearchPaths);
+        for (const auto& assetEntry : zoneDefinition.m_assets)
+        {
+            iwdWriter->AddFile(assetEntry.m_asset_name);
+        }
+
+        if (!iwdWriter->Write())
+        {
+            std::cout << "Writing iwd failed." << std::endl;
```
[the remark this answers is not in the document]
JezuzLizard: Fixed.

src/ObjWriting/ObjContainer/IWD/IWDWriter.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include <algorithm>
+#include <iostream>
+
+class IWDWriterImpl final : public IWDWriter
+{
+public:
+    explicit IWDWriterImpl(const std::string& iwdFileName, ISearchPath* assetSearchPath)
+        : m_filename(iwdFileName),
+          m_asset_search_path(assetSearchPath),
+          m_files(0)
```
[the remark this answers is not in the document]
JezuzLizard: Changed to use the clear() method.

src/ObjWriting/ObjContainer/IWD/IWDWriter.cpp
```diff
@@ -0,0 +1,106 @@
...
+        m_zipfile = zipOpen(m_filename.c_str(), APPEND_STATUS_CREATE);
+        if (!m_zipfile)
+        {
+            throw std::runtime_error("Error creating IWD file: \"" + m_filename + "\"");
+        }
+    }
+
+    void AddFile(std::string fileName) override
+    {
+        m_files.emplace_back(std::move(fileName));
```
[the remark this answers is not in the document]
Laupetin: This is not true. The `std::move` does make the `emplace_back` use the move constructor. That is necessary because calling the method itself does make the copy already since it's not a reference.

So it's correct to either use `std::move` here or otherwise make the parameter a `const std::string&` and use no `std::move`

src/ObjWriting/ObjContainer/IWD/IWDWriter.cpp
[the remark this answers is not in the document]
Laupetin: You should not need to explicitly initialize a vector. If you just omit any initialization it will call the default constructor of `std::vector` which does properly initialize it as empty vector

src/ObjWriting/ObjContainer/IWD/IWDWriter.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+class IWDWriterImpl final : public IWDWriter
+{
+public:
+    explicit IWDWriterImpl(const std::string& iwdFileName, ISearchPath* assetSearchPath)
+        : m_filename(iwdFileName),
+          m_asset_search_path(assetSearchPath),
+          m_files(0)
+    {
+        m_zipfile = zipOpen(m_filename.c_str(), APPEND_STATUS_CREATE);
```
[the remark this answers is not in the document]
Laupetin: As far as i understood it, this does make exactly what you are expecting: It creates the file instead of appending.

The `APPEND_STATUS_` is just the prefix and the `CREATE` is the actual chosen value.

Compared to the source:
```cpp
#define APPEND_STATUS_CREATE        (0)
#define APPEND_STATUS_CREATEAFTER   (1)
#define APPEND_STATUS_ADDINZIP      (2)
```

src/ObjWriting/ObjContainer/IWD/IWDWriter.cpp
Laupetin: I mean what's wrong with it. I think i used `all_of` a fair bit in the code :sweat_smile:

src/Linker/Linker.cpp
```diff
@@ -45,15 +46,11 @@ enum class ProjectType
...
+    IWD,
     MAX
 };
 
-constexpr const char* PROJECT_TYPE_NAMES[static_cast<unsigned>(ProjectType::MAX)]{
-    "none",
-    "fastfile",
-    "ipak",
-};
+constexpr const char* PROJECT_TYPE_NAMES[static_cast<std::underlying_type_t<ProjectType>>(ProjectType::MAX)]{"none", "fastfile", "ipak", "iwd"};
```
Laupetin, line 53: Unfortunately clang-format does not behave particularly cool when it comes to array initializer lists.

You can force it to make way less awkward formatting by adding a trailing comma `,` after the last entry which is what i do everywhere where i catch it until they introduce a better way to handle this :sweat_smile:

src/Linker/Linker.cpp
```diff
@@ -462,6 +459,32 @@ class LinkerImpl final : public Linker
...
+        auto iwdFilePath(iwdFolderPath);
+        const auto iwdFileName = zoneDefinition.m_name + ".iwd";
+        iwdFilePath.append(iwdFileName);
+
+        fs::create_directories(iwdFolderPath);
+
+        const auto iwdWriter = IWDWriter::Create(iwdFilePath.string(), &assetSearchPaths);
+        for (const auto& assetEntry : zoneDefinition.m_assets)
+        {
+            iwdWriter->AddFile(assetEntry.m_asset_name);
```
Laupetin, line 474: Does that mean it does try to write every asset into the IWD with the asset name interpreted as the full path?

That would mean that you cannot just simply use the same zone files as for the fastfile, right?
Because in a zone file for a fastfile you would write:
```
image,test_image
```
and it would refer to `images/test_image.iwi`.

In the context of the IWD here it would probably throw a "File not found"?

src/Linker/Linker.cpp
```diff
@@ -462,6 +459,32 @@ class LinkerImpl final : public Linker
...
+        }
+
+        if (!iwdWriter->Write())
+        {
+            std::cerr << "Writing iwd failed." << std::endl;
+            return false;
+        }
+
+        std::cout << "Created iwd \"" << iwdFilePath.string() << "\"\n";
```
Laupetin, line 483: Is there a reason for mixing `std::endl` and `\n`? :sweat_smile: 

I know i did it but i try to stick to just `\n` atm since `std::endl` also flushes and that can cost a bit of performance.

src/ObjWriting/ObjContainer/IWD/IWDWriter.cpp
```diff
@@ -0,0 +1,121 @@
+#include "IWDWriter.h"
+
+#include "../minizip/zip.h"
```
Laupetin, line 3: Please don't use relative paths in includes :sweat_smile: If minizip is currently not in the include paths for "ObjWriting" you may need to add it in the premake script "ObjWriting.lua"

src/ObjWriting/ObjContainer/IWD/IWDWriter.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+class IWDWriterImpl final : public IWDWriter
+{
+public:
+    explicit IWDWriterImpl(const std::string& iwdFileName, ISearchPath* assetSearchPath)
+        : m_filename(iwdFileName),
+          m_asset_search_path(assetSearchPath)
+    {
+        m_files.clear();
+        m_added_files.clear();
```
Laupetin: Clearing vectors in constructors is not necessary because the vector will be initialize by the automatically called default constructor of `std::vector`

src/ObjWriting/ObjContainer/IWD/IWDWriter.cpp
diamante0018: No, I mean std::all_of(m_files.cbegin(), m_files.cend()); constant iterator since the structure is unchanged.

src/ObjWriting/ObjContainer/IWD/IWDWriter.cpp
```diff
@@ -0,0 +1,121 @@
...
+        m_zipfile = zipOpen(m_filename.c_str(), APPEND_STATUS_CREATE);
+        if (!m_zipfile)
+        {
+            throw std::runtime_error("Error creating IWD file: \"" + m_filename + "\"");
```
Laupetin: Would be good to log this on std::cerr and have a graceful termination with non-zero exit code.
If you just throw an `std::runtime_error` this will unfortunately not be logged and will just crash.

And not being able to create a file may happen for example when having an invalid character in the path name.

src/ObjWriting/ObjContainer/IWD/IWDWriter.cpp
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    void AddFile(std::string fileName) override
+    {
+        const auto existingEntry = std::find(m_added_files.cbegin(), m_added_files.cend(), fileName);
+        if (existingEntry != m_added_files.cend())
+        {
+            std::cerr << "Duplicate entry in " << m_filename << "'s zone definition; ignoring" << "\n";
```
Laupetin: you can just make it `...ignoring\n";` you don't need the extra output command :stuck_out_tongue:

src/ObjWriting/ObjContainer/IWD/IWDWriter.cpp
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    void AddFileToArchive(const std::string& fileName, const std::unique_ptr<char[]>& fileData, const size_t& fileSize)
+    {
+        zip_fileinfo zi = {};
+
+        if (zipOpenNewFileInZip(m_zipfile, fileName.c_str(), &zi, NULL, 0, NULL, 0, NULL, Z_DEFLATED, Z_DEFAULT_COMPRESSION) != ZIP_OK)
+        {
+            zipClose(m_zipfile, NULL);
+            throw std::runtime_error("Could not open new file in IWD \"" + fileName + "\"\n");
```
Laupetin: Please don't use `std::runtime_error` :sweat_smile: This does not log anything and will just crash and not graceful exit

src/ObjWriting/ObjContainer/IWD/IWDWriter.cpp
```diff
@@ -0,0 +1,121 @@
...
+        zipClose(m_zipfile, NULL);
+        return result;
+    }
+
+private:
+    const std::string m_filename;
+
+    ISearchPath* m_asset_search_path;
+    std::vector<std::string> m_files;
+    std::vector<std::string> m_added_files;
```
Laupetin: If you use an [`std::unsorted_set`](https://en.cppreference.com/w/cpp/container/unordered_set) for this, this will not have to iterate through every file and do a basic string comparison for each previously added file on adding a file.

Sets make use of hashing and are therefore magnitudes faster for this.

src/ObjWriting/ObjContainer/IWD/IWDWriter.cpp
Laupetin: Okay i understand, fair

src/Linker/Linker.cpp
JezuzLizard: Hmm that is true. Would that mean IWDs should have their own type of definition that is only paths to files? Because you can plausibly load any kind of asset from raw in games such as CoD4/T4. I suppose it could be handled by making it so each asset type that doesn't have the full path in its name has a folder: for example image asset to images folder this way you wouldn't need to specify the images folder in the file name.

Laupetin: I mean while it's true that the older games support a lot of assets from IWDs, they should be consistent in what assets they put in which files, right? For example MW2 does only use IWDs for iwi images and streamed sounds.

I think some older titles also have weapon files in IWDs.

The way i'd do it is just have as much as possible in fastfiles and not in IWDs. At least for the first iteration.

Or what do you think?

Laupetin: Maybe I should add: The idea i had in mind on how to handle these other project types in the future was not to write different zone files necessarly.

I added a thing called "assetlists" for IPaks previously for example.
In the future i thought it might be neat to be able to refer to an assetlist of the written fastfile to just "put everything from the fastfile to an IWD or IPak".

Like Linker would link a fastfile mod.ff and create an assetlists/mod.csv file which can then be refered to from the IWD zone file with `assetlist,mod`.

That would require the way you define a zone and the way you define an IWD to match though :sweat_smile:

PR #71, opened 2023-12-25
feat: add warning for dupe localize assets

### LINKED ISSUE #70: Issue a warning when localized strings are duplicated
If OAT encounters a duplicated localized string it should issue a warning to make sure the user is aware that one version of a localized string is overwritten by another occurrence of it.

Otherwise there may be confusion when one occurrence of the string is changed by no change happens in the final output.

diamante0018:
Closes #70


src/ObjLoading/Parsing/LocalizeFile/Sequence/SequenceLocalizeFileLanguageValue.cpp
```diff
@@ -31,5 +31,13 @@ void SequenceLocalizeFileLanguageValue::ProcessMatch(LocalizeFileParserState* st
     state->m_current_reference_languages.emplace(langName);
 
     if (langName == state->m_language_name_caps)
-        state->m_entries.emplace_back(state->m_current_reference, valueToken.StringValue());
+    {
+        const auto& currentReference = state->m_current_reference;
+        if (const auto i = state->m_entries.find(currentReference); i != state->m_entries.end())
+        {
+            std::cout << "Localize: Reference \"" << currentReference << "\" was already defined. Using new value \"" << valueToken.StringValue() << "\"\n";
```
Laupetin, line 38: I think logging the value is a bit too much :sweat_smile: Probably too verbose since it can also be multi-line and everything.
The main point is to inform the user that there is a duplicate key so i guess just including the reference should be good enough :stuck_out_tongue:

src/ObjLoading/Parsing/LocalizeFile/LocalizeFileParserState.h
```diff
@@ -2,15 +2,15 @@
...
 #include <unordered_set>
-#include <vector>
 
 class LocalizeFileParserState
 {
 public:
     bool m_end;
 
-    std::vector<LocalizeFileEntry> m_entries;
+    std::map<std::string, std::string> m_entries;
```
Laupetin, line 13: After you removed the `LocalizeFileEntry` struct here i think you removed all usages of it. You can probably remove it then.

src/ObjLoading/Parsing/LocalizeFile/Sequence/SequenceLocalizeFileLanguageValue.cpp
```diff
@@ -31,5 +31,13 @@ void SequenceLocalizeFileLanguageValue::ProcessMatch(LocalizeFileParserState* st
     state->m_current_reference_languages.emplace(langName);
 
     if (langName == state->m_language_name_caps)
-        state->m_entries.emplace_back(state->m_current_reference, valueToken.StringValue());
+    {
+        const auto& currentReference = state->m_current_reference;
+        if (const auto i = state->m_entries.find(currentReference); i != state->m_entries.end())
```
Laupetin, line 36: The state is only for the current parsing session.
Meaning this works for duplicate entries in the same file.
When you have two separate str files that contain the same keys, it will not be noticed however.

My suggestion to solve this is to have a ZoneAssetLoaderState.
There is an interface called `IZoneAssetLoaderState`. If you inherit from it you can create a class that can be instanced once per zone and can hold a state across multiple asset loading calls per zone.

To retrieve it, you can do:
```
manager->GetAssetLoadingContext()->GetZoneAssetLoaderState<LocalizeZoneState>();
```
in any AssetLoader's `LoadFromRaw` class.
It will give you a state of the given class or will construct it if it was not created yet (needs to have a public default constructor).

So you could for example make a `LocalizeZoneState` class and initialize the `LocalizeFileParserState` with a pointer to it.
The zone state could hold a set of all existing keys in the zone.

diamante0018:
I think this is too complicated for me to do. I might not be able to finish this PR.

Laupetin:
> I think this is too complicated for me to do. I might not be able to finish this PR.

I can add a commit doing that so you can see what I mean. If that's okay for you

diamante0018:
That's okay. Please do. I address the other two comments but I can't
understand the latter.

diamante0018:
Oops, I forgot to push the two comments I said I addressed. I can do that later

diamante0018:
I pushed a new commit which addresses only the two minor comments, however, I will await your advice about for big change you requested.

src/ObjLoading/Parsing/LocalizeFile/Sequence/SequenceLocalizeFileLanguageValue.cpp
Laupetin: Pushed a commit to address this.

Issue #26, opened 2023-10-21
[T6] Sound linking/dumping


JezuzLizard:
Currently sounds cannot be added or modified using OAT because the game uses a special storage structure for sounds. Streamed sounds are stored in .sabs extension files and loaded sounds are stored in .sabl extension files. In order to allow for making mods that have custom or modified sounds the sabs/sabl structures need to be reversed. Additionally the sound/snddriverglobals/sndpatch assets need to be dumpable and linkable to allow modifying the sound aliases which the engine uses to know how to play the sound.

So what needs to be done is the following:

- sabs/sabl dumping
- sabs/sabl writing
- sound/snddriverglobals/sndpatch asset dumping
- sound/snddriverglobals/sndpatch asset writing

https://github.com/master131/Black-Ops-II-Sound-Studio This repo has useful research about the sabs/sabl structures and can already export sounds from a sabs/sabl file.

JezuzLizard:
Closing this as it was completed by a linked pull request.

PR #99, opened 2024-01-26
[T6] - SndBank Linking and Sound Bank (.sabs/.sabl) Writing

skiff:
Issue: #26 

Properly link SndBank assets for T6. This includes generating the appropriate sound bank file needed for loaded and streamed sounds. Supports linking the sound aliases, sound reverbs and sound ducks.

Note: The Sound Bank files have a checksum in them that is supposed to be an MD5 hash of the data section. However, the game does not validate this in any way. All that is checked is that the same checksum exists in the actual SndBank asset. Since T6 is the only supported game that uses sound bank files, it is not necessary to calculate this and instead the SndBank asset and Sound Bank file are populated with default ('\xCC') data for the checksum. If other games are added in the future that utilize the Sound Bank files, this may need to be corrected


skiff:
I think there may be an issue somewhere still with truly custom sounds. Need to investigate

skiff:
Fixed the issues with fully custom sounds. Custom sabl and sabs files are loading correctly now

src/ObjCommon/Csv/ParsedCsv.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+#include <Csv/CsvStream.h>
```
Laupetin, line 2: Project internal headers should use quotes `#include ""`

src/ObjLoading/ObjContainer/SoundBank/SoundBankWriter.cpp
```diff
@@ -0,0 +1,285 @@
...
+            Write(PAD_DATA.data(), writeSize);
+
+            paddingSizeLeft -= writeSize;
+        }
+    }
+
+    void AlignToChunk()
+    {
+        if ((m_current_offset & 0xF) != 0)
+            Pad(0x10 - (m_current_offset & 0xF));
```
Laupetin, line 80:
```cpp
if (m_current_offset % 16 != 0)
    Pad(m_current_offset % 16);
```
I think that's more readable 😅

src/ObjLoading/ObjContainer/SoundBank/SoundBankWriter.cpp
```diff
@@ -0,0 +1,285 @@
...
+          m_current_offset(0),
+          m_total_size(0),
+          m_entry_section_offset(0),
+          m_checksum_section_offset(0)
+    {
+    }
+
+    void AddSound(const std::string& soundFilePath, unsigned int soundId, bool looping, bool streamed) override
+    {
+        this->m_sounds.push_back({soundFilePath, soundId, looping, streamed});
```
Laupetin, line 50:
```cpp
this->m_sounds.emplace_back(soundFilePath, soundId, looping, streamed);
```

src/ObjLoading/Game/T6/AssetLoaders/AssetLoaderSoundBank.cpp
```diff
@@ -0,0 +1,584 @@
...
+            continue;
+
+        auto idx = sndBank->alias[i].id % sndBank->aliasCount;
+        while (sndBank->aliasIndex[idx].next != USHRT_MAX)
+        {
+            idx = sndBank->aliasIndex[idx].next;
+        }
+
+        auto offset = 1u;
+        auto freeIdx = USHRT_MAX;
```
Laupetin, line 237: careful: `freeIdx` has type `int` here. Because USHRT_MAX is just a macro for `0xffff` which makes this an int.
If you want to do the same thing the c++ way while also being guaranteed to get the right type you can use:
```cpp
auto freeIdx = std::numeric_limits<unsigned short>::max();
```

(It's a constexpr method so it will be optimized out, dont worry)

Laupetin:
Thanks for making this feature :) The CSV Reader for headers definitely is a good additional. Also nice branding section 😄

src/ObjCommon/Game/T6/ObjConstantsT6.h
```diff
@@ -24,5 +26,14 @@ namespace T6
         static constexpr const char* GDF_FILENAME_WEAPON_ATTACHMENT = "attachment.gdf";
         static constexpr const char* GDF_FILENAME_WEAPON_ATTACHMENT_UNIQUE = "attachmentunique.gdf";
         static constexpr const char* GDF_FILENAME_ZBARRIER = "zbarrier.gdf";
+
+        static const std::array<std::string, 27> SOUND_GROUPS;
```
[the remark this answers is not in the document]
skiff: The reason I wrote it this way is from a code readability standpoint, I felt it was a lot easier to read `SOUND_GROUPS.size()` rather than `std::extent_v<decltype(SOUND_GROUPS)>`. When building the sound aliases I have to reference these lists and their size a lot. I didn't want to use a vector since its all constants, but this felt like an okay in between. I can swap it back if you'd prefer

src/ObjLoading/Game/T6/AssetLoaders/AssetLoaderSoundBank.cpp
```diff
@@ -0,0 +1,584 @@
...
+        {
+            std::cerr << "Loaded Sound Bank for " << assetName << " failed to generate. Please check your build files." << std::endl;
+            return false;
+        }
+    }
+
+    // write the output streamed sound bank
+    if (sabsWriter)
+    {
+        auto size = static_cast<size_t>(sabsWriter->Write());
```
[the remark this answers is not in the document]
skiff: I agree the way that its organized currently doesn't make sense. I was originally planning to do something similar to what you suggested -- create a structure that while linking a zone would hold all of the sound data that needs to be written. Then after writing the zone out, the ZoneWriter could also write out all of the SoundBanks. My biggest concern for that was I didn't really like holding all the sound file data in memory if it wasn't necessary. It's probably not an issue since I think you already do something similar with the dumping and for images with ipaks so it would probably be okay. I can work on re-organizing now or after merged, whatever works best for you

src/ObjLoading/ObjContainer/SoundBank/SoundBankWriter.cpp
skiff: I changed it to use mod, but the example you gave does not do the same thing without the "0x10 - " in the pad call

src/ObjLoading/ObjContainer/SoundBank/SoundBankWriter.cpp
skiff: Changing to this breaks the build

src/ObjLoading/ObjContainer/SoundBank/SoundBankWriter.cpp
Laupetin: yeah you're right, messed up the code suggestion 👍

src/ObjLoading/ObjContainer/SoundBank/SoundBankWriter.cpp
Laupetin: Ahh okay you made this a struct without an explicit constructor. I usually try to avoid that and use classes with constructors. That makes sure that there cant be some odd behaviour with std::string not having its constructor called and makes sure that the members have default values set.

If you do this:
```cpp
class SoundBankEntryInfo
{
public:
	SoundBankEntryInfo()
		: m_sound_id(0u),
		  m_looping(false),
		  m_streamed(false)
	{
	}

	SoundBankEntryInfo(std::string filePath, const unsigned int soundId, const bool looping, const bool streamed)
		: m_file_path(std::move(filePath)),
		  m_sound_id(soundId),
		  m_looping(looping),
		  m_streamed(streamed)
	{
	}

	std::string m_file_path;
	unsigned int m_sound_id;
	bool m_looping;
	bool m_streamed;
};
```
That suggestion will work, avoids a copy of the path string (i think it currently does one, not 100% on that one though) and is more in-line with how the rest of the codebase works

src/ObjLoading/Game/T6/AssetLoaders/AssetLoaderSoundBank.cpp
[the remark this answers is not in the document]
Laupetin: It will be changed somewhen else. I think this should just be merged for now since people can get value out of it, whether it is refactored or not.

PR #125, opened 2024-02-24
do not add duplicate sounds to the sab files

skiff:
Some sound aliases use the same asset file. We do not need to add the same file to a bank multiple times


Laupetin (approved):
👍

PR #132, opened 2024-03-03
maint(premake5): enable /LARGEADDRESSAWARE at least on Windows

### LINKED ISSUE #115: Make all executables large address aware
Without setting the `IMAGE_FILE_LARGE_ADDRESS_AWARE` flag a 32bit application will only be able to allocate as much as 2GB memory ([see also](https://learn.microsoft.com/en-us/windows/win32/memory/memory-limits-for-windows-releases)).

Since both Unlinker and Linker can load any amount of fastfiles beforehand, setting this option will increase the amount of possible loaded fastfiles by a lot.
It's not really an urgent issue but there should not be a downside to it really. 
With premake this is probably done by using this: https://premake.github.io/docs/largeaddressaware/

diamante0018:
Closes #115

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -14,6 +14,7 @@ workspace "OpenAssetTools"
     symbols "On"
     systemversion "latest"
     cppdialect "C++20"
+    largeaddressaware "on"
 
 	flags {
 		"MultiProcessorCompile"


diamante0018:
image

PR #181, opened 2024-04-28
feature: dump leaderboard definitions on IW4/IW5

diamante0018:
Tested with: `-v "D:\Steam\steamapps\common\Call of Duty Modern Warfare 3\zone\english\code_post_gfx_mp.ff"`


src/ObjCommon/Game/IW5/Json/JsonLeaderboardDef.h
```diff
@@ -0,0 +1,99 @@
...
+                                     {LBCOL_TYPE_PRESTIGE,  "prestige" },
+                                     {LBCOL_TYPE_BIGNUMBER, "bignumber"},
+                                     {LBCOL_TYPE_PERCENT,   "percent"  },
+                                     {LBCOL_TYPE_TIME_FULL, "time_full"},
+    });
+
+    NLOHMANN_JSON_SERIALIZE_ENUM(LbAggType,
+                                 {
+                                     {LBAGG_TYPE_MIN,  "normal"},
+                                     {LBAGG_TYPE_MAX,  "rank"  },
```
Laupetin, line 29: These two seem wrong

src/ObjCommon/Game/IW5/Json/JsonLeaderboardDef.h
```diff
@@ -0,0 +1,99 @@
...
+                                 {
+                                     {LBUPDATE_TYPE_NORMAL,  "normal" },
+                                     {LBUPDATE_TYPE_RANK,    "rank"   },
+                                     {LBUPDATE_TYPE_COMBINE, "combine"},
+    });
+
+    class JsonColumnDef
+    {
+    public:
+        std::optional<std::string> name;
```
Laupetin, line 44: My guess would be that name should never be non-existant, no?

src/ObjCommon/Game/IW5/Json/JsonLeaderboardDef.h
```diff
@@ -0,0 +1,99 @@
...
+            in.at("id").get_to(out.id);
+            in.at("propertyId").get_to(out.propertyId);
+            in.at("hidden").get_to(out.hidden);
+            optional_from_json(in, "statName", out.statName);
+            in.at("type").get_to(out.type);
+            in.at("precision").get_to(out.precision);
+            in.at("agg").get_to(out.agg);
+            in.at("uiCalColX").get_to(out.uiCalColX);
+            in.at("uiCalColY").get_to(out.uiCalColY);
+        };
```
Laupetin, line 81: Why is this here? That should already be done by the macro below?

src/ObjCommon/Game/IW5/Json/JsonLeaderboardDef.h
```diff
@@ -0,0 +1,99 @@
...
+    };
+
+    NLOHMANN_DEFINE_TYPE_EXTENSION(JsonColumnDef, name, id, propertyId, hidden, statName, type, agg, uiCalColX, uiCalColY);
+
+    class JsonLeaderboardDef
+    {
+    public:
+        std::optional<std::string> name;
```
Laupetin, line 89: The name of the asset should already be given by the filename and specified in the `.zone`. Having it in the file again makes the value be specified in two places which makes it annoying and unintuitive to change.

src/ObjWriting/Game/IW5/AssetDumpers/AssetDumperLeaderboardDef.cpp
```diff
@@ -0,0 +1,39 @@
...
+    {
+        std::ranges::replace(sanitizedFileName, '*', '_');
+        const auto parenthesisPos = sanitizedFileName.find('(');
+        if (parenthesisPos != std::string::npos)
+            sanitizedFileName.erase(parenthesisPos);
+        sanitizedFileName = "generated/" + sanitizedFileName;
+    }
+
+    return std::format("leaderboards/{}.json", sanitizedFileName);
+}
```
Laupetin, line 24: This will not be necessary, it only was for materials because in T6 Radiant generated materials can have `*` and `()` in them.

You can just use `std::format("leaderboards/{}.json")` and inline that where the function was used

src/ObjWriting/Game/IW5/LeaderboardDef/JsonLeaderboardDefWriter.cpp
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        void Dump(const LeaderboardDef* leaderboardDef)
+        {
+            JsonLeaderboardDef jsonLeaderboardDef;
+            CreateJsonLeaderboardDef(jsonLeaderboardDef, *leaderboardDef);
+
+            json jRoot = jsonLeaderboardDef;
+
+            jRoot["_type"] = "leaderboardDef";
```
Laupetin, line 29: i'd just call the asset "leaderboard", thats also what the `zone` file does

src/ObjCommon/Game/IW5/Json/JsonLeaderboardDef.h
```diff
@@ -0,0 +1,99 @@
...
+        int id;
+        int xpColId;
+        int prestigeColId;
+        std::vector<JsonColumnDef> columns;
+        LbUpdateType updateType;
+        int trackTypes;
```
Laupetin, line 95: Did you investigate whether this is related to either `mp/trackercolumns.csv` or `trackerColType_e`?

src/ObjCommon/Game/IW5/Json/JsonLeaderboardDef.h
```diff
@@ -0,0 +1,99 @@
...
+
+    NLOHMANN_DEFINE_TYPE_EXTENSION(JsonColumnDef, name, id, propertyId, hidden, statName, type, agg, uiCalColX, uiCalColY);
+
+    class JsonLeaderboardDef
+    {
+    public:
+        std::optional<std::string> name;
+        int id;
+        int xpColId;
+        int prestigeColId;
```
Laupetin, line 92: do you know what these col ids are? are they an enum value as well?

src/ObjWriting/Game/IW5/LeaderboardDef/JsonLeaderboardDefWriter.cpp
```diff
@@ -0,0 +1,97 @@
...
+                jColumnDef.name = AssetName(lbColumnDef.name);
+            }
+
+            jColumnDef.id = lbColumnDef.id;
+            jColumnDef.propertyId = lbColumnDef.propertyId;
+            jColumnDef.hidden = lbColumnDef.hidden;
+
+            if (lbColumnDef.statName)
+            {
+                jColumnDef.statName = AssetName(lbColumnDef.statName);
```
Laupetin, line 57: `statName` is not an asset name 😅 This function is just meant to resolve asset references. Same goes for `name` but that should not be a property anyway, see above.

In fact the AssetName function should not be used anywhere here, since the Leaderboard does not reference any other assets.

src/ObjCommon/Game/IW5/Json/JsonLeaderboardDef.h
diamante0018: I think this enum could be used in the propertyId field for the column not the actual leaderboard def. Not 100% sure

diamante0018: I don't think it's mp/trackercolumns either because the value always stays about 1 or 0

diamante0018: I am unable to determine at this time so I will not serialize this field as an enum I think

src/ObjCommon/Game/IW5/Json/JsonLeaderboardDef.h
diamante0018: I have no idea

src/ObjCommon/Game/IW5/Json/JsonLeaderboardDef.h
diamante0018: I  don't think this is that enum. If the dumping is correct the 63 value means it's not really that enum or part of that CSV.
image

src/ObjCommon/Game/IW5/Json/JsonLeaderboardDef.h
diamante0018: Could not find any enums that would explain this field either being 1 or -1

diamante0018:
I was bored and I also did IW4 dumping

`-v "D:\Steam\steamapps\common\Call of Duty Modern Warfare 2\zone\english\code_post_gfx_mp.ff"` only has one leader board though. Perhaps I should find out if there are more files with this.

diamante0018:
image
I guess it's not really used on IW4. I could try to download the playlist from their CDN because IW4x has that off by default

src/ObjWriting/Game/IW4/LeaderboardDef/JsonLeaderboardDefWriter.cpp
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        void Dump(const LeaderboardDef* leaderboardDef)
+        {
+            JsonLeaderboardDef jsonLeaderboardDef;
+            CreateJsonLeaderboardDef(jsonLeaderboardDef, *leaderboardDef);
+
+            json jRoot = jsonLeaderboardDef;
+
+            jRoot["name"] = leaderboardDef->name;
```
Laupetin, line 29: Name of the leaderboard should not be inside the file as previously commented 😅

src/ObjWriting/Game/IW5/LeaderboardDef/JsonLeaderboardDefWriter.cpp
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        void Dump(const LeaderboardDef* leaderboardDef)
+        {
+            JsonLeaderboardDef jsonLeaderboardDef;
+            CreateJsonLeaderboardDef(jsonLeaderboardDef, *leaderboardDef);
+
+            json jRoot = jsonLeaderboardDef;
+
+            jRoot["name"] = leaderboardDef->name;
```
Laupetin, line 29: Name of the leaderboard should not be inside the file as previously commented 😅

src/Common/Game/IW5/IW5_Assets.h
```diff
@@ -4380,6 +4380,19 @@ namespace IW5
...
+    {
+        TRK_ALLTIME = 0x0,
+        TRK_WEEKLY = 0x1,
+        TRK_MONTHLY = 0x2,
+        TRK_PRESTIGE_ALLTIME = 0x3,
+        TRK_PRESTIGE_WEEKLY = 0x4,
+        TRK_PRESTIGE_MONTHLY = 0x5,
+        TRK_DAILY = 0x6,
+        TRK_PRESTIGE_DAILY = 0x7,
+        TRK_COUNT = 0x8,
```
Laupetin, line 4393: i try to not hardcode the count value but let the compiler select it itself, so adding an entry only adds a line but not change the count property and prevents accidents in not increasing it.

src/ObjCommon/Game/IW4/Json/JsonLeaderboardDef.h
```diff
@@ -0,0 +1,58 @@
...
+        int id;
+        int propertyId;
+        bool hidden;
+        std::string statName;
+        LbColType type;
+        int precision;
+        LbAggType agg;
+    };
+
+    NLOHMANN_DEFINE_TYPE_EXTENSION(JsonColumnDef, name, id, propertyId, hidden, statName, type, agg);
```
Laupetin, line 46: Missing the `precision` property

PR #182, opened 2024-04-30
feat: dump xmodels as gltf/glb

Laupetin:
Dumping XModels as gltf makes sure they can be read by most 3D modelling software without additional addons.
It is also significantly faster than dumping as OBJ or XMODEL_EXPORT.

Exemplary comparison of time it took me to dump all models from IW4's common_mp.ff

| Format | Time |
| -- | -- |
| XMODEL_EXPORT | 1m22.631s |
| OBJ    |       45.756s |
| GLTF  |        7.437s |
| GLB   |        5.106s |

* Add support for dumping xmodels in all games as GLTF (blob included as base64)
* Add support for dumping xmodels in all games as GLB (binary version of GLTF)
* Change default model format to GLB
* Rework OBJ dumping to make use of XModel common structs to reduce code duplication

TODO:
- [x] Specify max and min for gltf POSITION accessor
- [x] Include bone weights in gltf files
- [x] Improve material information included in gltf files


Laupetin:
Seems to work fine for all games now

Issue #184, opened 2024-05-06
[Linker]: Unable to extend patch_mp on latest and older versions


diamante0018:
Using  v0.12.1:
`.\bin\Unlinker.exe "D:\Steam\steamapps\common\Call of Duty Modern Warfare 3\zone\english\patch_mp.ff"`

copying over the dumped files to the correct location to proceed with re-linking (<OAT  Dir>/zone_raw).

`.\bin\Linker.exe -l "D:\Steam\steamapps\common\Call of Duty Modern Warfare 3\zone\english\patch_mp.ff" --menu-permissive --menu-no-optimization patch_mp`

Menu flags are required

The error list is HUGE

[oat.txt](https://github.com/Laupetin/OpenAssetTools/files/15219774/oat.txt)

I tried on a release version before the asset names were "lower cased" but it's the same issue, the Linker cannot find those xanims even with the asset names being half lowercase half uppercase.

Laupetin:
This should now be fixed with #188 being merged

diamante0018:
I honestly can't tell because it's either this fast file as genuinely some big issues or OAT can't find some assets in memory.

On a stock untouched raw dump of the fast file:
` -l "D:\Steam\steamapps\common\Call of Duty Modern Warfare 3\zone\english\patch_mp.ff" --menu-permissive --menu-no-optimization patch_mp`
[oat-log1.txt](https://github.com/Laupetin/OpenAssetTools/files/15285480/oat-log1.txt)

When actually extending the fast file with a mod
` -l "D:\Steam\steamapps\common\Call of Duty Modern Warfare 3\zone\english\patch_mp.ff" -l "mod.ff" --menu-permissive --menu-no-optimization patch_mp`

[oat-log2.txt](https://github.com/Laupetin/OpenAssetTools/files/15285495/oat-log2.txt)

Laupetin:
> I honestly can't tell because it's either this fast file as genuinely some big issues or OAT can't find some assets in memory.
> 
> On a stock untouched raw dump of the fast file: ` -l "D:\Steam\steamapps\common\Call of Duty Modern Warfare 3\zone\english\patch_mp.ff" --menu-permissive --menu-no-optimization patch_mp` [oat-log1.txt](https://github.com/Laupetin/OpenAssetTools/files/15285480/oat-log1.txt)
> 
[...]

Issue #180, opened 2024-04-28
Asset lowercase conversion does not work with IW5 leaderboard assets


Laupetin:
IW5's code_post_gfx_mp contains leaderboard assets that have an all uppercase name (like `DEFAULT_LB`).
This unfortunately means you currently cannot load them from in-memory since all attempts to load an asset with name `DEFAULT_LB` will result in OAT trying to find an asset with the name `default_lb`, which it cannot find.

The best solution here is probably to normalize asset names when adding them to the asset pool, since the game itself does compare and hash asset names case insensitive.

Laupetin:
This should now be fixed with #188 being merged

diamante0018:
I can confirm the issue was fixed.

Laupetin:
Neat

PR #188, opened 2024-05-11
feat: make asset lookup work case insensitively entirely

Laupetin:
This normalizes all asset names before adding them to the asset lookup, enabling case insensitive search for assets.

### DIFF
--- a/src/ZoneCommon/Pool/XAssetInfo.h
+++ b/src/ZoneCommon/Pool/XAssetInfo.h
@@ -30,14 +30,6 @@ template<> struct std::hash<IndirectAssetReference>
 class XAssetInfoGeneric
 {
 public:
-    asset_type_t m_type;
-    std::string m_name;
-    void* m_ptr;
-    std::vector<XAssetInfoGeneric*> m_dependencies;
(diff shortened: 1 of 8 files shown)


src/ZoneCommon/Pool/XAssetInfo.h
```diff
@@ -61,6 +53,16 @@ class XAssetInfoGeneric
     XAssetInfoGeneric(XAssetInfoGeneric&& other) noexcept = default;
     XAssetInfoGeneric& operator=(const XAssetInfoGeneric& other) = default;
     XAssetInfoGeneric& operator=(XAssetInfoGeneric&& other) noexcept = default;
+
+    static std::string NormalizeAssetName(std::string input);
```
Laupetin, line 57: Localize assets seem to be dumped lowercase now which is not intended

Laupetin: I ended up not normalizing the name of the asset in the XAssetInfo, so this is solved now

Laupetin:
This should address #184 and #180

PR #191, opened 2024-05-12
feature: dump leaderboard definitions on T6

diamante0018:
`-v "D:\Steam\steamapps\common\Call of Duty Black Ops II\zone\all\patch_mp.ff"`

There seems to be no correlation between the column count and the dw column count.


Laupetin (approved):
Added two commits ontop to do similar changes as to the IW4/IW5 PR. Thank you for the contributions :)

PR #205, opened 2024-05-25
feat(iw4): load leaderboard definitions from raw

diamante0018:


src/ObjLoading/Game/IW4/AssetLoaders/AssetLoaderLeaderboard.cpp
```diff
@@ -15,3 +18,31 @@ void* AssetLoaderLeaderboard::CreateEmptyAsset(const std::string& assetName, Mem
...
+
+bool AssetLoaderLeaderboard::CanLoadFromRaw() const
+{
+    return true;
+}
+
+std::string AssetLoaderLeaderboard::GetFileNameForAsset(const std::string& assetName)
+{
+    return std::format("leaderboards/{}.json", assetName);
+}
```
Laupetin, line 30: There are some things that are done in some places in the codebase that i would not do anymore that way:
This is one thing: I currently would not put these kinds of methods in the AssetLoaderLeaderboard class.

This is because this requires this method to be declared in the header, even though it is not used outside of the cpp file at all.
My current approach is to define these kinds of methods in an anonymous namespace in the cpp file (anonymous namespace because this avoids conflicts with other methods with that same name when linking).

In this case it would probably also make sense to omit the method entirely because its only used in one place and its a one liner, so you dont really reduce complexity by introducing a method.
The "GetFileNameForAsset" methods were mainly introduced to extract code when generating the filenames requires more logic.

All in all a lot of the stuff im going to comment is not anything tragic at all but i thought i'd just clearify my thoughts on these places since im currently doing it differently for new code :P

src/ObjLoading/Game/IW4/AssetLoaders/AssetLoaderLeaderboard.cpp
```diff
@@ -15,3 +18,31 @@ void* AssetLoaderLeaderboard::CreateEmptyAsset(const std::string& assetName, Mem
...
+    if (!file.IsOpen())
+        return false;
+
+    auto* leaderboardDef = memory->Alloc<LeaderboardDef>();
+    leaderboardDef->name = memory->Dup(assetName.c_str());
+
+    if (LoadLeaderboardAsJson(*file.m_stream, *leaderboardDef, memory))
+        manager->AddAsset<AssetLeaderboard>(assetName, leaderboardDef);
+    else
+        std::cerr << "Failed to load leaderboard \"" << assetName << "\"\n";
```
Laupetin, line 45: Since the codebase is now C++20 i'm now trying to make use of `std::format` in these case as much as possible.
For these kinds of places this also solves the issue of sometimes experiencing disconnected messages were one part of a log line would be behind another line (like you experienced in your runner).

It's also good preparation for a proper logging system that i kind of want to introduce at some point 😅

src/ObjCommon/Game/IW4/Leaderboard/JsonLeaderboardDef.h
```diff
@@ -12,6 +12,8 @@
 
 namespace IW4
 {
+    static constexpr auto LEADERBOARD_DEF_VERSION = 1u;
```
Laupetin, line 15: i dont think its necessary to introduce a constant in ObjCommon for this. The reason is that the loader will probably have to support multiple versions for backwards compatibility while the dumper will only dump the latest version.
So these two parts are kind of not necessarly connected in that regard.

src/ObjLoading/Game/IW4/Leaderboard/JsonLeaderboardDefLoader.cpp
```diff
@@ -0,0 +1,110 @@
...
+            return CreateLeaderboardFromJson(jLeaderboard, leaderboardDef);
+        }
+
+    private:
+        bool CreateColumnDefFromJson(const JsonColumnDef& jColumn, LbColumnDef& lbColumnDef, LeaderboardDef& leaderboardDef) const
+        {
+            lbColumnDef.name = m_memory.Dup(jColumn.name.c_str());
+
+            if (jColumn.propertyId)
+                lbColumnDef.propertyId = jColumn.propertyId.value();
```
Laupetin, line 49: It's kind of easier to understand if you set an explicit `else` branch to set a default value, can you maybe do that? 😅 I know the struct is zero initialized by the new allocation code but it requires thinking about it for a bit and setting a default value is still required if it does not correspond to zero memory.

src/ObjLoading/Game/IW4/Leaderboard/JsonLeaderboardDefLoader.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+            if (jColumn.hidden)
+                lbColumnDef.hidden = jColumn.hidden.value();
+
+            if (jColumn.statName)
+                lbColumnDef.statName = m_memory.Dup(jColumn.statName.value().c_str());
```
Laupetin, line 55:
```suggestion
                lbColumnDef.statName = m_memory.Dup(jColumn.statName->c_str());
```

You can use `->` for std::optional, which does the same as `.value()`

src/ObjLoading/Game/IW4/Leaderboard/JsonLeaderboardDefLoader.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+        bool CreateLeaderboardFromJson(const JsonLeaderboardDef& jLeaderboardDef, LeaderboardDef& leaderboardDef) const
+        {
+            leaderboardDef.id = jLeaderboardDef.id;
+
+            if (jLeaderboardDef.xpColId)
+                leaderboardDef.xpColId = jLeaderboardDef.xpColId.value();
+
+            if (jLeaderboardDef.prestigeColId)
+                leaderboardDef.prestigeColId = jLeaderboardDef.prestigeColId.value();
```
Laupetin, line 75: The colIds here are `-1` as its default value, so you will **have to** add `else` branches for default values here

Laupetin (commented):
Thanks for correcting my optional mistakes :)

src/ObjLoading/Game/IW4/Leaderboard/JsonLeaderboardDefLoader.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+            if (jLeaderboardDef.xpColId)
+                leaderboardDef.xpColId = jLeaderboardDef.xpColId.value();
+            else
+                leaderboardDef.xpColId = -1;
+
+            if (jLeaderboardDef.prestigeColId)
+                leaderboardDef.prestigeColId = jLeaderboardDef.prestigeColId.value();
+            else
+                leaderboardDef.prestigeColId = -1;
```
Laupetin, line 87: std::optional has a neat API that allows you to do this:
```suggestion
            leaderboardDef.xpColId = jLeaderboardDef.xpColId.value_or(-1);
            leaderboardDef.prestigeColId = jLeaderboardDef.prestigeColId.value_or(-1);
```

src/ObjLoading/Game/IW4/Leaderboard/JsonLeaderboardDefLoader.cpp
```diff
@@ -0,0 +1,122 @@
...
+                lbColumnDef.statName = m_memory.Dup(jColumn.statName->c_str());
+            else
+                lbColumnDef.statName = nullptr;
+
+            lbColumnDef.type = jColumn.type;
+
+            if (jColumn.precision)
+                lbColumnDef.precision = jColumn.precision.value();
+            else
+                lbColumnDef.precision = 0;
```
Laupetin, line 68: Same thing here:

```suggestion
            lbColumnDef.propertyId = jColumn.propertyId.value_or(0);
            lbColumnDef.hidden = jColumn.hidden.value_or(false);

            if (jColumn.statName)
                lbColumnDef.statName = m_memory.Dup(jColumn.statName->c_str());
            else
                lbColumnDef.statName = nullptr;

            lbColumnDef.type = jColumn.type;

            lbColumnDef.precision = jColumn.precision.value_or(0);
```

src/ObjLoading/Game/IW4/Leaderboard/JsonLeaderboardDefLoader.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+            const auto jLeaderboard = jRoot.get<JsonLeaderboardDef>();
+            return CreateLeaderboardFromJson(jLeaderboard, leaderboardDef);
+        }
+
+    private:
+        bool CreateColumnDefFromJson(const JsonColumnDef& jColumn, LbColumnDef& lbColumnDef, LeaderboardDef& leaderboardDef) const
+        {
+            lbColumnDef.name = m_memory.Dup(jColumn.name.c_str());
+
```
Laupetin, line 47: The column does not have its `id` field set from the `colId` json field

Laupetin (approved):

Laupetin:
I just fixed the two things myself :) But feel free to look at it, might be relevant for the other two games. Thanks for the contribution :)

PR #206, opened 2024-05-26
feat(iw5): load leaderboard definitions from raw

diamante0018:


src/ObjLoading/Game/IW5/Leaderboard/JsonLeaderboardDefLoader.cpp
```diff
@@ -0,0 +1,145 @@
...
+                gameFlags |= gameFlag;
+
+            return true;
+        }
+
+        bool CreateColumnDefFromJson(const JsonColumnDef& jColumn, LbColumnDef& lbColumnDef, LeaderboardDef& leaderboardDef) const
+        {
+            lbColumnDef.name = m_memory.Dup(jColumn.name.c_str());
+
+            if (jColumn.propertyId)
```
Laupetin, line 56: The column id is not set here 😅

src/ObjLoading/Game/IW5/Leaderboard/JsonLeaderboardDefLoader.cpp
```diff
@@ -0,0 +1,145 @@
...
+                lbColumnDef.propertyId = jColumn.propertyId.value();
```
Laupetin, line 57: For IW5 as well: You can use the std::optional value_or method :P
See IW4 PR

src/ObjLoading/Game/IW5/Leaderboard/JsonLeaderboardDefLoader.cpp
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+    private:
+        static bool CreateTrackTypeFlagsFromJson(const JsonLeaderboardDef& jLeaderboardDef, int& gameFlags)
+        {
+            for (const auto gameFlag : jLeaderboardDef.trackTypes)
+                gameFlags |= gameFlag;
+
+            return true;
+        }
```
Laupetin, line 50: I think you copied some code here but it doesn't apply here the same way 😅 
The naming is a bit off, its not gameFlags here, so the code is probably from materials.
The other thing is that the enum entries have normal enum values, not the flag values.
So you have to or with `1 << enumValue`, not the enum value.

Laupetin (approved):

Laupetin:
I quickly fixed the stuff here as well, i split it up into different commits if you want to take a look at it :) Thanks for the contribution. Whats missing is adding the stuff to the README but i think i'll do that after the PRs to not create merge conflicts.

PR #245, opened 2024-09-06
Added support for Unsigned Int asseccors

ElectroGamesDev:


ElectroGamesDev:
Here is an example of a model that had this issue: https://sketchfab.com/3d-models/apple-low-poly-59ec6f10d16f4fe9a683462e3c780386

src/ObjLoading/XModel/Gltf/Internal/GltfAccessor.cpp
```diff
@@ -435,3 +435,111 @@ bool UnsignedShortAccessor::GetUnsignedVec4(const size_t index, unsigned (&out)[
...
+}
+
+bool UnsignedIntAccessor::GetUnsignedVec4(const size_t index, unsigned (&out)[4]) const
+{
+    assert(m_type == JsonAccessorType::VEC4);
+    if (index >= m_count)
+        return false;
+
+    uint32_t temp[4];
+    if (!m_buffer_view->ReadElement(temp, index, sizeof(uint8_t[4]), m_byte_offset))
```
Laupetin, line 536:
```suggestion
    if (!m_buffer_view->ReadElement(temp, index, sizeof(uint32_t[4]), m_byte_offset))
```
It does read the wrong size here i think. You probably copied it from the uint16 accessor and it is also wrong there :sweat_smile: Can you adjust that one as well maybe?

Laupetin (approved):
Thank you for the contribution. The apple still does not work but thats my fault. I'll look further into it 😅

Issue #177, opened 2024-04-25
Feature request: in verbose mode print when asset parsing is done and zone writing begins


diamante0018:
When I use OAT in a workflow over at git.alterware we have a small and not-so-powerful machine doing the ci work.

so what ends up happening is when menu parsing is done it might stay "unresponsive" for several minutes because, in reality, it's writing the zone itself.

OAT prints a message when the zone is done creating, but if we are in verbose mode I'd like a message showing that zone creation has begun so I know it's not stuck parsing some menu file.

![image](https://github.com/Laupetin/OpenAssetTools/assets/37080671/eb2fec81-7a51-4d2b-817a-932aa395e907)

diamante0018:
I will build on this issue since they are all similar and request the following ideas coming from zonetool:

basic profiling trivia:

log to the stdout how much time it took to build a zone and how much memory it took:

https://git.alterware.dev/Future/zonetool/src/branch/develop/src/ZoneTool/ZoneTool.cpp#L285
https://git.alterware.dev/Future/zonetool/src/branch/develop/src/IW5/Zone.cpp#L145
https://git.alterware.dev/Future/zonetool/src/branch/develop/src/IW5/Zone.cpp#L272

PR #260, opened 2024-09-11
feat(Linker): add verbose loggin when writing of zones begins

diamante0018:
Tackles in part #177 
Tracking time spent while linking is cool but I could do that some other time and then I can close that issue.


diamante0018:
This  works as I wanted
image

Laupetin (approved):
Thanks for the addition :)

PR #261, opened 2024-09-12
feat: track time spent linking zone

### LINKED ISSUE #177: Feature request: in verbose mode print when asset parsing is done and zone writing begins

diamante0018:
Closes #177 

This completes the features I requested. This is greatly inspired by how ZoneTool tracks the msec passed when linking a Zone.

image

### DIFF
--- a/src/ZoneWriting/ZoneWriting.cpp
+++ b/src/ZoneWriting/ZoneWriting.cpp
@@ -8,6 +8,7 @@
 #include "Writing/IZoneWriterFactory.h"
 
 #include <chrono>
+#include <format>
 #include <iostream>
 
 IZoneWriterFactory* ZoneWriterFactories[]{
@@ -42,7 +43,7 @@ bool ZoneWriting::WriteZone(std::ostream& stream, Zone* zone)
 
     const auto end = std::chrono::high_resolution_clock::now();
 
-    std::cout << std::format("Writing zone \"{}\" took {} msec.\n", zone->m_name, std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count());
+    std::cout << std::format("Writing zone \"{}\" took {} ms.\n", zone->m_name, std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count());
 
     return result;
 }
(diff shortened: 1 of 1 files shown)


src/ZoneWriting/ZoneWriting.cpp
```diff
@@ -7,6 +7,9 @@
 #include "Game/T6/ZoneWriterFactoryT6.h"
 #include "Writing/IZoneWriterFactory.h"
 
+#include <chrono>
+#include <iostream>
```
Laupetin, line 11: Since you use `std::format`, please also include the `#include <format>` header

Laupetin (approved):
Other than the header, looks fine :) thanks

src/ZoneWriting/ZoneWriting.cpp
```diff
@@ -29,9 +34,15 @@ bool ZoneWriting::WriteZone(std::ostream& stream, Zone* zone)
...
+        std::cerr << std::format("Could not create ZoneWriter for zone \"{}\".\n", zone->m_name);
         return false;
     }
 
-    return zoneWriter->WriteZone(stream);
+    const auto result = zoneWriter->WriteZone(stream);
+
+    const auto end = std::chrono::high_resolution_clock::now();
+
+    std::cout << std::format("Writing zone \"{}\" took {} msec.\n", zone->m_name, std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count());
```
Laupetin, line 45: Oh also the other code refers to the time unit "milliseconds" as `ms`, not `msec`.
Its not that this is necessarly worse, but would be nice to be consistent in the codebase.

diamante0018:
Just saw the review. Will try to address it later today

PR #274, opened 2024-10-01
Dump and load correct volume and pitch values for T6 sound aliases

Jbleezy:
I became a scholar of music theory to figure out what these values were representing.

BO1 and BO3 sound aliases have volume values ranging from 0 to 100, and pitch values ranging from -2400 to 1200.

The amp ratio value is due to 100 decibels being equal to 100,000 amplitude, but the value is stored in 16 bits.

[Decibel converter](https://www.silisoftware.com/tools/db.php)

[Cents converter](https://sengpielaudio.com/calculator-centsratio.htm)

### DIFF
--- a/src/Common/Game/T6/CommonT6.h
+++ b/src/Common/Game/T6/CommonT6.h
@@ -3,15 +3,12 @@
 #include "T6.h"
 
 #include <cctype>
-#include <limits>
 
 namespace T6
 {
     class Common
     {
     public:
-        static constexpr double AMP_RATIO = 100000.0 / std::numeric_limits<uint16_t>::max();
-
         static constexpr int Com_HashKey(const char* str, const int maxLen)
         {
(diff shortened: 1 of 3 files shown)


Laupetin (approved):
Thanks for the PR :) Your guess about the input formats was very correct.
I reversed a bit of the T5 modding tools to find out the exact formulas used though, so i adjusted them a bit.
And made the code a bit better to read.
Thanks a lot :D

Laupetin:
Oh also: I will follow up with another Pull Request to refactor a good bit of the loading and dumping code.
Your changes are already backwards incompatible so i thought i could just as well add a couple more to not break existing csvs more than once.
It should fix a good amount of weird quirks, small mistakes and missing data, so afterwards it might be worth checking whether the odd sound issues are fixed.

Issue #155, opened 2024-04-05
Sound volumes are lower than expected


Jbleezy:
If you export a sound that is already in game using BO2 Sound Studio and use the same sound alias line that is given from Unlinker, the sound comes out lower than normal in game. This is especially noticeable with voicelines.

This could be an issue with exporting from BO2 Sound Studio, I'm not sure.

Sammy-0:
The variables in unlinked aliases are a bit weird too, like volume and pitch have super high values for some reason. In particular the LFE alias for weapons seems exaggerated compared to the base game; maybe it was broken in the OG game or something?

skiff:
@Jbleezy @Sammy-0 can you provide some examples? The fastfile and soundbank the original sound is in as well as an OAT built zone/soundbank that has this issue?

Jbleezy:
I was adding the Richtofen Grief voicelines to Nuketown when I really noticed this issue.
Fastfile: `en_common_zm.ff`
Soundbank: `zmb_common.english`
[sound_volume_test.zip](https://github.com/Laupetin/OpenAssetTools/files/14952410/sound_volume_test.zip)

Jbleezy:
In BO1 soundaliases, `vol_min` and `vol_max` go from 0 - 100, `pitch_min` goes from -100 - 0, and `pitch_max` goes from 0 - 100.

Jbleezy:
Here is a video showcasing the issue and comparing it to what the volume should be:

https://vimeo.com/1009355363

PR #277, opened 2024-10-06
refactor: sound bank loading dumping

Laupetin:
Adjust a lot of soundbank dumping and writing code.
Some of these changes are backwards incompatible, but they are needed for values that make more sense than previously
and also some fields were previously dumped and loaded incorrectly or were missing entirely.


Jbleezy:
Closes #155

PR #286, opened 2024-10-14
feat: improve building IWD and IPak obj containers

Laupetin:
## Previously

Currently building IPaks is pretty annoying and building IWDs is not implemented yet.
Previously OBJ containers had to be built as seperate targets.
If they were supposed to include obj data for assets from the built zone they had to include an assetlist which have not been generated automatically and have a fully defined zone file.

```
// zone_raw/mod/mod.zone
build,mod_ff
build,mod_ipak
```
```
// zone_raw/mod/mod_ff.zone
>game,T6
>name,mod
>type,fastfile
>level.ipak_read,mod
include,content
```
```
// zone_raw/mod/mod_ipak.zone
>game,T6
>name,mod
>type,ipak
include,content
```
```
// zone_raw/mod/content.zone
image,image1
image,image2
xmodel,xmodel1
// ...
```

## Now

This PR tries to unify the building process of fastfile and additional obj containers into a single target.
This aims to improve building of IPaks, IWDs and in the future soundbanks.

```
// zone_raw/mod/mod.zone
>game,T6
>name,mod
>ipak,mymoddedipakname
>iwd,mymoddediwdname

image,image1
image,image2
xmodel,xmodel1
```

This is supposed to result in a built `mod.ff`, `mymoddedipakname.ipak` and `mymoddediwdname.iwd`.
The IPak and IWD contain all supported obj data of the assets that follow.

You can switch to another IWD or IPak by redefining `>ipak,newipak` or `>iwd,newiwd` respectively.
This will cause all following obj data to go into a new obj container.
This way you can split data.

Specifying the respective metadata key with an empty value `>ipak,` or `>iwd,` then you can stop writing data to this obj container.

Closes: #27


Laupetin:
Closes: #27

Laupetin:
It turned out that implementing these kinds of features required a bigger rewrite than i initially thought.
So I took the chance and restructured everything for a future separation of asset compilers and asset loaders.

I hope this is not breaking anything, i will test thoroughly and write more unit tests before a release i guess.

Laupetin:
This does not add support for ending containers early by omitting name.
The parser does not currently emit new line tokens which means that an omitted value would prefetch the next token which could be from the next line.

So this would require a bigger adaption to the ZoneDefinition parsing which i might only add after introducing ANTLR4..

PR #333, opened 2025-01-10
tiny mistakes in the filenames of the make scripts

awsms:
Couldn't find ./scripts/make_release.sh, since it was ./scripts/make-release.sh


Laupetin (approved):
Good catch, thanks :)

PR #391, opened 2025-03-24
Update T6 asset loading code to use ASSET_TYPE_CLIPMAP_PVS.

LJW-Dev:
BO2's ASSET_TYPE_CLIPMAP is unused, and instead uses ASSET_TYPE_CLIPMAP_PVS both in the game code and fastfiles, so I've updated the asset code to only use ASSET_TYPE_CLIPMAP_PVS. This doesn't effect the .zone files and is just a backend change.

### DIFF
--- a/src/ObjLoading/Game/T6/ObjLoaderT6.cpp
+++ b/src/ObjLoading/Game/T6/ObjLoaderT6.cpp
@@ -404,7 +404,6 @@ namespace T6
             collection.AddAssetCreator(CreateImageLoader(memory, searchPath));
             collection.AddAssetCreator(CreateSoundBankLoader(memory, searchPath));
             // collection.AddAssetCreator(std::make_unique<AssetLoaderSoundPatch>(memory));
-            // collection.AddAssetCreator(std::make_unique<AssetLoaderClipMap>(memory));
             // collection.AddAssetCreator(std::make_unique<AssetLoaderClipMapPvs>(memory));
             // collection.AddAssetCreator(std::make_unique<AssetLoaderComWorld>(memory));
             // collection.AddAssetCreator(std::make_unique<AssetLoaderGameWorldSp>(memory));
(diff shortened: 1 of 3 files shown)


Laupetin:
Hey there, thanks for the PR :) You're saying that the .zone files are not affected, they are though 😅
The asset name you are changing to `clipmap_unused` is being used in:
* dumping asset lists
* loading asset lists (which can be done from a .zone file)
* loading zone files resolved assets via this name

You called the branch "fix", do you have something specific in mind you want to fix with this?

LJW-Dev:
Sorry, I wasn't clear in my original comment. 
The issue was that there is two duplicate "clipmap" names in the `ASSET_TYPE_NAMES` array, and when resolving the asset type name from the asset type both `ASSET_TYPE_CLIPMAP` and `ASSET_TYPE_CLIPMAP_PVS` resolve to "clipmap". So, when linking a zone using a custom clipmap, it tries to resolve "clipmap" back to an asset type, and chooses `ASSET_TYPE_CLIPMAP` over `ASSET_TYPE_CLIPMAP_PVS` as it is the first in the list when resolving.

I've been looking into dumping and loading all the BSP info from maps to possibly create a map editing tool, and I found out that bo2 doesn't use `ASSET_TYPE_CLIPMAP` at all and only uses `ASSET_TYPE_CLIPMAP_PVS`. When trying to load a map using the non-PVS clipmap type it errors out and says it can't find the clipmap for the map, while if the PVS clipmap type is used it doesn't error and I can load the map.

This PR fixes this issue by renaming the unused clipmap asset name to "clipmap_unused", and this won't effect / will fix:
- dumping asset lists: clipmaps within original fastfiles use `ASSET_TYPE_CLIPMAP_PVS` so the asset type name will still be "clipmap".
- loading asset lists: This is only an issue when custom clipmaps are loaded, but OAT doesn't have this functionality yet (And this PR fixes the issue anyways).
- loading zone files resolved assets via this name: I found it was a simple fix when writing the clipmap loading/dumping code, when referencing the asset type I just had to be sure to use `ASSET_TYPE_CLIPMAP_PVS` instead of `ASSET_TYPE_CLIPMAP`.

Laupetin:
Okay i understand. Yeah that makes sense, good find, thanks :)

The CI currently does not run through due to a bug that was caused by a new compiler version.
I fixed it on the `main` branch, so if you rebase on my `main` branch it should run through and i can merge it.

Laupetin (dismissed):

LJW-Dev:
No worries, i'll do that now.

LJW-Dev:
My bad, i rebased with some extra code that I forgot to delete :sweat_smile: should be good to go now.

Laupetin:
thanks for the contribution :) I'll adapt IW3,IW4,T5 for it as well, there seems to be the same problem

PR #404, opened 2025-04-13
Added XModelBin Export

GoastcraftHD:
Tested on T5, T6 and IW4. Unfortunately I don't own IW3 and IW5, but these games should work just fine.

### DIFF
--- a/src/ObjWriting/XModel/Export/XModelBinWriter.cpp
+++ b/src/ObjWriting/XModel/Export/XModelBinWriter.cpp
@@ -8,11 +8,8 @@
 
 #include <chrono>
 #include <format>
-#include <iomanip>
-#include <iostream>
-#include <lib/lz4.h>
 #include <limits>
-#include <ostream>
+#include <lz4.h>
 #include <sstream>
 #include <string>
 
--- a/thirdparty/lz4.lua
+++ b/thirdparty/lz4.lua
@@ -3,7 +3,7 @@ lz4 = {}
 function lz4:include(includes)
 	if includes:handle(self:name()) then
 		includedirs {
-			path.join(ThirdPartyFolder(), "lz4")
+			path.join(ThirdPartyFolder(), "lz4/lib")
 		}
 	end
 end


.gitmodules
```diff
@@ -16,3 +16,6 @@
 [submodule "thirdparty/eigen"]
 	path = thirdparty/eigen
 	url = https://gitlab.com/libeigen/eigen.git
+[submodule "thirdparty/lz4"]
+	path = thirdparty/lz4
+	url = git@github.com:lz4/lz4.git
```
diamante0018, line 21: this url should use the https link like the others, not everyone has ssh setup

Laupetin:
Hey there, thanks for the PR :)

Since adding a new export format adds complexity and potential sources of bugs, can you maybe give me a bit of context on the feature? 😅 
I have not yet needed `xmodel_bin` support so knowing your usecase would help me a bit on evaluating, especially since the amount of tools that support this format is probably comparatively narrow.

* What benefits does supporting `xmodel_bin` grant over the existing formats? Does it have advantages over the existing `xmodel_export` other than being smaller in size?
* Do you use specific tooling that works better with it?
* Is there some specific use you have for models in this format?
* Is there any specification that you referenced for implementation?

I'm also asking because even though I implemented `xmodel_export` first, I kind of shifted my focus away from it due to the poor tooling support in my toolchain and not being an industry standard format.
So I did not and were not planning on adding support for reading xmodels from that format for now and concentrate on making gltf support better.

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,474 @@
...
+        }
+    }
+
+    void WriteAlignedString(const std::string& string)
+    {
+        const auto paddingSize = ((string.size() + 1 + 0x3) & 0xFFFFFFFFFFFFFC) - (string.size() + 1);
+        const auto padding = std::make_unique<uint8_t[]>(static_cast<size_t>(paddingSize));
+
+        m_writer.WriteNullTerminatedString(string);
+        m_writer.Write(reinterpret_cast<uint8_t*>(padding.get()), static_cast<uint32_t>(paddingSize));
```
Laupetin, line 106: What you are doing is you are writing a zero terminated string with padding to the next multiple of 4.
You are allocating new uninitialized memory for this and write it directly without initialization?

That seems slow and may leak previously allocated memory values.
Padding can at max be 3 bytes long.

```cpp
static constexpr uint8_t PADDING[4] = {0u, 0u, 0u, 0u};
const auto paddingSize = (4u - (string.size() + 1u) % 4u) % 4u;

m_writer.WriteNullTerminatedString(string);
m_writer.Write(PADDING, paddingSize);
```

(Btw there is a utility method `Utils/Alignment.h => Align` for aligning to next multiple of in case you need it)

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,474 @@
...
+
+    void WriteUInt16(const int16_t hash, const uint16_t value)
+    {
+        m_writer.Write(hash);
+        m_writer.Write(value);
+    }
+
+    int16_t ClampFloatToShort(const float value) const
+    {
+        return std::clamp(static_cast<int16_t>(32767 * std::clamp(value, -1.0f, 1.0f)), static_cast<int16_t>(-32767), static_cast<int16_t>(32767));
```
Laupetin, line 129: Can you maybe use `std::numeric_limits<int16_t>::max()` and `::min()` respectively?

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    uint8_t ClampFloatToUByte(const float value) const
+    {
+        return std::clamp(static_cast<uint8_t>(255 * std::clamp(value, -1.0f, 1.0f)), static_cast<uint8_t>(0), static_cast<uint8_t>(255));
```
Laupetin, line 134: And `std::numeric_limits<uint8_t>::max()` here?

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    void WriteMaterials(const XModelCommon& xmodel)
+    {
+        WriteInt16(static_cast<int16_t>(XModelBinHash::MATERIAL_COUNT), static_cast<int16_t>(xmodel.m_materials.size()));
+
+        size_t materialNum = 0u;
+        for (const auto& material : xmodel.m_materials)
+        {
+            const auto colorMapPath = "../images/" + material.colorMapName + ".dds";
```
Laupetin, line 366: `std::format("../images/{}.dds", material.colorMapName)`

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,474 @@
...
+        auto estimatedCompressedFileSize = LZ4_compressBound(m_writer.GetPosition());
+        auto compressedBuffer = new char[estimatedCompressedFileSize];
+        auto actualCompressedFileSize =
+            LZ4_compress_default(reinterpret_cast<const char*>(m_writer.Data()), compressedBuffer, m_writer.GetPosition(), estimatedCompressedFileSize);
+
+        uint64_t uncompressedSize = m_writer.GetPosition();
+        char uncompressedSizeChar[4];
+        std::memcpy(uncompressedSizeChar, &uncompressedSize, sizeof(uncompressedSizeChar));
+
+        const char magic[5] = {0x2A, 0x4C, 0x5A, 0x34, 0x2A}; // *LZ4*
```
Laupetin, line 461:
```cpp
static constexpr char MAGIC[5] = {'*', 'L', 'Z', '4', '*'};
```
Why not write the values directly? 😅

Laupetin:
I tested it with Blender 2.9 and the plugin from CoDEmanX and SE2Dev and it works pretty well :) Had a few comments, most of them related to your style being more close to C than I do in this project (not necessarly a bad thing, just trying to keep it more consistent to the rest of the code base 😅).

GoastcraftHD:
Thanks for taking your time to explain everything to me. Since I'm just a junior I still have a lot to learn.

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+#pragma warning(push, 0)
+#include <Eigen>
+#pragma warning(pop)
+
+#include <chrono>
+#include <iomanip>
+#include <iostream>
+#include <lib/lz4.h>
+#include <ostream>
```
Laupetin, line 11: You are using `std::ostringstream` which is defined in header `sstream`.
It is possible to sometimes use the standard lib features without including the appropriate headers because other headers that you included do include them.

These inclusions of your included headers may change though through changing unrelated header files, the standard lib version being used (version updates; MSVC STL vs glibc).

So to be safe across platforms and versions it is a good idea to always include all headers of all things you are making use of.

So i'd add the following headers:
```cpp
#include <format>
#include <limits>
#include <sstream>
#include <string>
```
These are the ones that I found being used in the file.

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,484 @@
...
+    uint8_t ClampFloatToUByte(const float value) const
+    {
+        return std::clamp(static_cast<uint8_t>(std::numeric_limits<uint8_t>::max() * std::clamp(value, -1.0f, 1.0f)),
+                          std::numeric_limits<uint8_t>::min(),
+                          std::numeric_limits<uint8_t>::max());
+    }
+
+    void WriteHeader(int16_t version)
+    {
+        WriteComment("OpenAssetTools XMODEL_BIN File");
```
Laupetin, line 151: Including branding is nice here :) I usually would make it include the version number as well for the following reason:

When somebody reports a bug related to a file being wrong, not being able to read it or something similar, it is then possible to see which version this file was generated with.

You can do that like this:
```cpp
#include "GitVersion.h"
// ...
WriteComment("OpenAssetTools " GIT_VERSION " XMODEL_BIN file");
```
The GitVersion.h file is generated by premake inside the build directory.

In C and C++, strings are concatinated statically when you place them next to each other without an operator.
So `"Hello" "World"` would result in `"HelloWorld"`.
That's how this works.

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+    XModelBinWriterBase(std::ostream& stream, std::string gameName, std::string zoneName)
+        : m_stream(stream),
+          m_game_name(std::move(gameName)),
+          m_zone_name(std::move(zoneName))
+    {
+    }
+
+    std::ostringstream m_writer;
+    std::ostream& m_stream;
```
Laupetin, line 220: The `m_stream` variable is not used in the parent class, consider moving it to the child class.

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,484 @@
...
+        PrepareVertexMerger(xmodel);
+        WriteHeader(7);
+        WriteBones(xmodel);
+        WriteVertices(xmodel);
+        WriteFaces(xmodel);
+        WriteObjects(xmodel);
+        WriteMaterials(xmodel);
+
+        auto estimatedCompressedFileSize = LZ4_compressBound(m_writer.str().size());
+        auto compressedBuffer = new char[estimatedCompressedFileSize];
```
Laupetin, line 464: You are leaking memory here, you are doing `new`, but not `delete`.

Best do
```cpp
const auto compressedBuffer = std::make_unique<char[]>(estimatedCompressedFileSize);
```
here, then the allocated buffer is automatically freed when the compressedBuffer variable leaves scope.

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,484 @@
...
+        auto actualCompressedFileSize = LZ4_compress_default(m_writer.str().c_str(), compressedBuffer, m_writer.str().size(), estimatedCompressedFileSize);
+
+        auto uncompressedSize = m_writer.str().size();
+        char uncompressedSizeChar[4];
+        std::memcpy(uncompressedSizeChar, &uncompressedSize, sizeof(uncompressedSizeChar));
```
Laupetin, line 469: Doing this is a bit dangerous here, the returned `size()` is a `size_t` which is 4 bytes in 32bit but 8bytes in 64bit.
Since the variable is `auto` type, it will get bigger here as well.

Additionally, you dont need another variable here since you can cast the pointer to the numeric variable.

```cpp
auto uncompressedSize = static_cast<uint32_t>(m_writer.str().size());
// ....
m_stream.write(reinterpret_cast<char*>(&uncompressedSize), sizeof(uncompressedSize));
```

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,484 @@
...
+    void WriteMaterials(const XModelCommon& xmodel)
+    {
+        WriteInt16(static_cast<int16_t>(XModelBinHash::MATERIAL_COUNT), static_cast<int16_t>(xmodel.m_materials.size()));
+
+        size_t materialNum = 0u;
+        for (const auto& material : xmodel.m_materials)
+        {
+            const auto colorMapPath = std::format("../images/{}.dds", material.colorMapName);
+
+            WriteInt16(static_cast<int16_t>(XModelBinHash::MATERIAL), static_cast<int16_t>(materialNum));
```
Laupetin, line 379: `WriteUInt16`?

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,484 @@
...
+        m_writer.write(reinterpret_cast<const char*>(PADDING), paddingSize);
+    }
+
+    void WriteComment(const std::string& comment)
+    {
+        Write(XModelBinHash::COMMENT);
+        WriteAlignedString(comment);
+    }
+
+    void WriteInt16(const int16_t hash, const int16_t value)
```
Laupetin, line 123: You made the hash enum be a `uint32_t` and the method here (and all the other methods that take this hash) use `int16_t`.
Why don't you make the enum and the method use `uint16_t`?

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,484 @@
...
+                Write(-1);
+
+            WriteAlignedString(bone.name);
+            boneNum++;
+        }
+
+        boneNum = 0;
+        for (const auto& bone : xmodel.m_bones)
+        {
+            WriteInt16(static_cast<int16_t>(XModelBinHash::BONE_INDEX), boneNum);
```
Laupetin, line 179: Why not `WriteUInt16` and making the variable be unsigned, as the boneNumber should not become negative?

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,484 @@
...
+    VertexMerger m_vertex_merger;
+};
+
+class XModelBinWriter7 final : public XModelBinWriterBase
+{
+    void WriteVertices(const XModelCommon& xmodel)
+    {
+        const auto& distinctVertexValues = m_vertex_merger.GetDistinctValues();
+
+        if (distinctVertexValues.size() > UINT16_MAX)
```
Laupetin, line 232: `UINT16_MAX` is the C way, `std::numeric_limits<uint16_t>::max()` the C++ way.

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,484 @@
...
+                // Use 16 bit
+                WriteUInt16(static_cast<int16_t>(XModelBinHash::VERT16), static_cast<uint16_t>(vertexNum));
+            }
+
+            XModelBinWriterBase::Write(XModelBinHash::OFFSET);
+            XModelBinWriterBase::Write(vertexPos.x);
+            XModelBinWriterBase::Write(vertexPos.y);
+            XModelBinWriterBase::Write(vertexPos.z);
+
+            WriteInt16(static_cast<int16_t>(XModelBinHash::VERT_WEIGHT_COUNT), static_cast<int16_t>(vertexPos.weightCount));
```
Laupetin, line 264: Why write a signed number here?

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,484 @@
...
+                }
+            }
+
+            objectIndex++;
+        }
+    }
+
+    void WriteObjects(const XModelCommon& xmodel)
+    {
+        WriteInt16(XModelBinHash::OBJECT_COUNT, static_cast<int16_t>(xmodel.m_objects.size()));
```
Laupetin, line 357: `WriteUInt16`?

src/ObjWriting/XModel/Export/XModelBinWriter.cpp
```diff
@@ -0,0 +1,484 @@
...
+            XModelBinWriterBase::Write(static_cast<int16_t>(objectNum));
+            WriteAlignedString(object.name);
+
+            objectNum++;
+        }
+    }
+
+    void WriteMaterials(const XModelCommon& xmodel)
+    {
+        WriteInt16(static_cast<int16_t>(XModelBinHash::MATERIAL_COUNT), static_cast<int16_t>(xmodel.m_materials.size()));
```
Laupetin, line 372: `WriteUInt16`?

Laupetin (commented):
Noticed a few more things :sweat_smile: Sorry.
Should be ready to merge afterwards though :+1:

Laupetin (approved):
Thank you for your PR :)

PR #409, opened 2025-04-20
chore: update Premake5

diamante0018:
`gmake2` deprecated -> `gmake`


diamante0018:
> eh, that's a breaking change. `premake5.lua` does have a check for premake version which will then have to be updated to at least `beta5`, since thats where the breaking change happened.
> 
> Also my distro did not update from beta4 yet 😅 Will merge then i guess 😬

Got it, updating required min version now

PR #411, opened 2025-04-20
fix: wrong "bin" directory when using executable from $PATH

### LINKED ISSUE #410: Linker does not recognize "bin" directory when being called from $PATH
Linking when Linker binary is not in an expected location (see [diamante0018/setup-oat](https://github.com/diamante0018/setup-oat)) may cause C++ exceptions (regression didn't happen between 0.8.3 through 0.19.something I think)

Image

diamante0018:
see: Closes #410 

Closes #410


diamante0018:
works for me on windows

PR #422, opened 2025-04-26
refactor: fix compilation warnings for x64

Laupetin:
Mostly usages of size_t which changes size to the word size which needs explicit conversions in x64.


src/Cryptography/Algorithms/AlgorithmMd5.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include "IHashFunction.h"
+
+#include <memory>
+
+namespace cryptography
+{
+    std::unique_ptr<IHashFunction> CreateMd5();
+}
```
diamante0018, line 10: Fatal  missing `// crytopgraphy `, clang-format missing. entire PR  invalidated
image

Laupetin: All cpp files are clang-format checked and my IDE auto formats on save. So if there is no comment then there shouldnt be :P

See ShortNamespaceLines on https://clang.llvm.org/docs/ClangFormatStyleOptions.html

PR #441, opened 2025-05-30
Fix BO2 IWI files saving cube textures incorrectly.

LJW-Dev:
When dumping cube images as an IWI, they are incorrectly saved with a 2D image header instead which makes it impossible to load them again (e.g. bo2 reflection probes). 

As the class TextureCube extends Texture2D, a dynamic_cast to Texture2D will work on both TextureCube and Texture2D. The IWI saving code checks a dynamic_cast to Texture2D first, so if a TextureCube class is given it will match Texture2D and be dumped as a Texture2D. This pull request fixes the issue by checking Texture2D after TextureCube so cube images are correctly dumped as a TextureCube.

This issue might be present in other IWI versions but I haven't checked.


Laupetin:
I added a fix for the other iwi format versions on top

PR #539, opened 2025-10-11
WIP t6 custom map loading

LJW-Dev:
Notes:
- Techsets are loaded and dumped with my custom implementation, I think it should use the official techset structure but this isn't super important.
- The world asset uses a .fbx file and uses OpenFBX as it's loader
- It doesn't use any xmodels apart from the skybox, as I found them too complicated to load from the world asset. Definitely should be implemented somehow though.
- You need 6 gsc files for the link to be successful and for the map to load on bo2, but 4 of them are redundant (the _fx and _amb GSC files). This could be changed to make it simpler for users to create their custom maps, but i'm not sure if simplifying it is a good design choice.
- Maps require footstep tables, but there's no loading/dumping of them right now. I add empty footstep tables so the map will load, we could change this so users can edit the footstep tables but i don't know how this would be done.


src/ObjLoading/Game/T6/CustomMap/BinarySpacePartitionTreePreCalc.h
```diff
@@ -0,0 +1,232 @@
...
+    }
+};
+
+union u_BSPNode;
+class BSPTree;
+
+class BSPLeaf
+{
+private:
+    std::vector<Object*> objectList;
```
Laupetin: That should probably be a unique_ptr

src/ObjLoading/Game/T6/CustomMap/ProjectCreator.cpp
```diff
@@ -0,0 +1,395 @@
...
+        num_vertices = ufbx_generate_indices(streams, 1, indices, num_indices, NULL, NULL);
+        _ASSERT(num_vertices != 0);
+
+        vertexVec.insert(vertexVec.end(), &vertices[0], &vertices[num_vertices]);
+
+        size_t currIndexVecSize = indexVec.size();
+        indexVec.resize(indexVec.size() + num_indices);
+        for (size_t m = 0; m < num_indices; m++)
+        {
+            indexVec[currIndexVecSize + m] = (uint16_t)indices[m];
```
Laupetin: nit: the code mixes C style casts and C++ style casts a lot. The project generally only uses C++ style casts to be more explicit with what should happen and prevent unintended sideeffects.

Laupetin (commented):
Hey there, thank you for the PR, the feature is pretty awesome :)
I've looked through a lot of the code by now, sorry it takes so long, it's a big PR 😅 
I probably won't have that much time in the upcoming days, so i'll submit the comments I have currently.

I haven't looked at CustomMapLinker.h yet, though since I do not know that much about the BSPs yet, I probably won't be able to say about that part 😅 
So generally my comments are more related to code structure and memory management for example.

Also please forgive that I'm adding a lot of comments, it's just my style of doing code review (I'm adding a comment when I have any special thoughts about some place so it ends up being pretty verbose) and does not mean that necessarly everything needs to be addressed or must be solved in a particular way necessarly.
Also I can address a lot of the stuff as well, you don't have to do everything 😛

Again, thanks for putting the work in :D Looking forward to this feature and how it evolves

src/Common/Game/T6/T6.h
```diff
@@ -158,6 +158,100 @@ namespace T6
...
+    {
+        std::string name;
+        std::string bspName;
+
+        customMapGfx gfxInfo;
+        customMapGfx colInfo;
+        
+        size_t modelCount;
+        customMapModel* models;
+    };
```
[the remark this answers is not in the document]
LJW-Dev: Would the separate header file be in the objcommon project or the objloading project?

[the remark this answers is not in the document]
Laupetin: Generally at the same location as the cpp file.
And for those -> if it is used in the loading code only -> ObjLoading
If the dumping code uses it as well -> ObjCommon

src/Common/Game/T6/T6_Assets.h
```diff
@@ -1278,6 +1278,30 @@ namespace T6
...
+
+
+    struct GfxLightGridRow
+    {
+        unsigned __int16 colStart;
+        unsigned __int16 colCount;
+        unsigned __int16 zStart;
+        unsigned __int16 zCount;
+        unsigned int firstEntry;
+        GfxLightGridUnk unk;
```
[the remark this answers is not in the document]
LJW-Dev: True, I can see it being a lookup table. I'll update the code to try match this better.

src/ObjLoading/Game/T6/BSP/Linker/BSPLinker.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include "../BSP.h"
```
Laupetin: The project does not use backwards directory traversal for paths 😅 

`Game/T6/BSP/BSP.h` is prefered

src/ObjLoading/Game/T6/BSP/LoaderBSP_T6.cpp
Laupetin: The rest of the files that are game specific do not have an underscore before the game suffix.

I guess you made it because otherwise you would not have a separator between the T of T6 and the P of BSP?
In that case maybe `LoaderBspT6` would be an option?

src/ObjLoading/Game/T6/BSP/BSPUtil.cpp
```diff
@@ -0,0 +1,173 @@
...
+        vec3_t result;
+        result.x = (maxs.x - mins.x) * 0.5f;
+        result.y = (maxs.y - mins.y) * 0.5f;
+        result.z = (maxs.z - mins.z) * 0.5f;
+        return result;
+    }
+
+    size_t BSPUtil::allignBy128(size_t size)
+    {
+        return ((size + 127) & 0xFFFFFF80);
```
Laupetin: There's a utility method for it already in `Utils/Alignment.h`:

```cpp
utils::Align(size, 128uz);
```

src/ObjLoading/Game/T6/BSP/BSPUtil.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include "BSPUtil.h"
+
+#include <cmath>
+#include <format>
+
+namespace BSP
+{
+    std::string BSPUtil::getFileNameForBSPAsset(std::string& assetName)
+    {
+        return std::format("BSP/{}", assetName);
```
Laupetin: Folders are generally lowercase. On Windows this does not make a difference, on Linux it does.

src/ObjLoading/Game/T6/BSP/BSPUtil.cpp
```diff
@@ -0,0 +1,173 @@
...
+            AABBMins.y = newAABBMins.y;
+
+        if (newAABBMaxs.y > AABBMaxs.y)
+            AABBMaxs.y = newAABBMaxs.y;
+
+        if (AABBMins.z > newAABBMins.z)
+            AABBMins.z = newAABBMins.z;
+
+        if (newAABBMaxs.z > AABBMaxs.z)
+            AABBMaxs.z = newAABBMaxs.z;
```
Laupetin: These can be changed to
```cpp
        AABBMins.x = std::min(AABBMins.x, newAABBMins.x);
        AABBMaxs.x = std::max(AABBMaxs.x, newAABBMaxs.x);

        AABBMins.y = std::min(AABBMins.y, newAABBMins.y);
        AABBMaxs.y = std::max(AABBMaxs.y, newAABBMaxs.y);

        AABBMins.z = std::min(AABBMins.z, newAABBMins.z);
        AABBMaxs.z = std::max(AABBMaxs.z, newAABBMaxs.z);
```

which is a bit more readable imo 😅 

Adjusted in a8456f88a49fae340446833234f33df1c2274e5b

src/ObjLoading/Game/T6/BSP/LoaderBSP_T6.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include "Asset/IAssetCreator.h"
+#include "Game/T6/T6.h"
```
Laupetin: nit: This header seems unused

adjusted in a8fc10df74a0acb264894edbf7988fde88ea0d85

src/ObjLoading/Game/T6/BSP/BSPCreator.cpp
```diff
@@ -0,0 +1,276 @@
+#include "BSPCreator.h"
+
+#include "BSPUtil.h"
+
+#include <ufbx.h>
+
+namespace
+{
+    using namespace BSP;
```
Laupetin: Usually you only do `using namespace` on global level since this technically embed one namespace into another which can get a little weird, especially if that namespace it is embedded into is named 😅 

adjusted in 08a8a5cd077ae217b1abeac4b0c8889f2b46d8b2

src/ObjLoading/Game/T6/BSP/LoaderBSP_T6.cpp
```diff
@@ -0,0 +1,60 @@
...
+    {
+    public:
+        BSPLoader(MemoryManager& memory, ISearchPath& searchPath, Zone& zone)
+            : m_memory(memory),
+              m_search_path(searchPath),
+              m_zone(zone)
+        {
+        }
+
+        std::optional<asset_type_t> GetHandlingAssetType() const override
```
Laupetin: nit: Not _that_ important but can be helpful sometimes: There is a c++ feature called [`nodiscard`](https://en.cppreference.com/w/cpp/language/attributes/nodiscard) that raises a compiler warning if the return value of a function is not used.

Helpful at functions that do nothing other than return a value.
Generally mostly applicable when a method is static or const.

adjusted in: 8f08aca15a2c13438e26cfe25e8aa17087ee6e3a

src/ObjLoading/Game/T6/BSP/Linker/BSPLinker.cpp
```diff
@@ -0,0 +1,106 @@
...
+            return false;
+        if (m_context.LoadDependency<AssetScript>("maps/mp/" + bsp->name + "_fx.gsc") == nullptr)
+            return false;
+
+        if (m_context.LoadDependency<AssetScript>("clientscripts/mp/" + bsp->name + ".csc") == nullptr)
+            return false;
+        if (m_context.LoadDependency<AssetScript>("clientscripts/mp/" + bsp->name + "_amb.csc") == nullptr)
+            return false;
+        if (m_context.LoadDependency<AssetScript>("clientscripts/mp/" + bsp->name + "_fx.csc") == nullptr)
+            return false;
```
Laupetin: nit: the codebase generally tries to use std::format instead of string concatination

adjusted in: 5e06df6286f30de54027f115c08c94e986bfe73c

src/ObjLoading/Game/T6/BSP/Linker/GfxWorldLinker.cpp
```diff
@@ -0,0 +1,772 @@
...
+        gfxWorld->outdoorLookupMatrix[0].x = xRecip;
+        gfxWorld->outdoorLookupMatrix[1].y = yRecip;
+        gfxWorld->outdoorLookupMatrix[2].z = zRecip;
+        gfxWorld->outdoorLookupMatrix[3].x = xScale;
+        gfxWorld->outdoorLookupMatrix[3].y = yScale;
+        gfxWorld->outdoorLookupMatrix[3].z = zScale;
+        gfxWorld->outdoorLookupMatrix[3].w = 1.0f;
+
+        std::string outdoorImageName = std::string("$outdoor");
+        auto outdoorImageAsset = m_context.LoadDependency<AssetImage>(outdoorImageName);
```
Laupetin: nit: since `LoadDependency` takes in a `const std::string&`, you can pass in a literal value directly instead of having to make a temporary variable that you only use once.

adjusted in: 049fa2b8ae999aabfe5f91355ce4b8a8eb692f6c

src/ObjLoading/Game/T6/BSP/BSP.h
```diff
@@ -0,0 +1,186 @@
...
+
+        BSPWorld gfxWorld;
+        BSPWorld colWorld;
+    };
+
+    // BSPGameConstants:
+    // These values are hardcoded ingame and will break the map if they are changed
+    namespace BSPGameConstants
+    {
+        constexpr unsigned int MAX_COLLISION_VERTS = UINT16_MAX;
```
Laupetin: nit: when there is a c++ way of doing things, i often prefer it over the C way 😅 

Numeric limits in c++ are defined via [`std::numeric_limits`](https://en.cppreference.com/w/cpp/types/numeric_limits.html).
While it may have the small benefit of always being typed correctly by default, it is mostly a personal taste thing.

adjusted in: 27c3c7dccb1069d7a7e9d26645261cb5d9c0482d

src/ObjLoading/Game/T6/BSP/BSPUtil.h
```diff
@@ -0,0 +1,24 @@
...
+        static void updateAABBWithPoint(vec3_t& point, vec3_t& AABBMins, vec3_t& AABBMaxs);
+        static vec3_t calcMiddleOfAABB(vec3_t& mins, vec3_t& maxs);
+        static vec3_t calcHalfSizeOfAABB(vec3_t& mins, vec3_t& maxs);
+        static size_t allignBy128(size_t size);
+        static float distBetweenPoints(vec3_t& p1, vec3_t& p2);
+        static void convertAnglesToAxis(vec3_t* angles, vec3_t* axis);
+        static void matrixTranspose3x3(const vec3_t* in, vec3_t* out);
+        static vec3_t convertStringToVec3(std::string& str);
+        static std::string convertVec3ToString(vec3_t& vec);
+    };
```
Laupetin: As I mentioned in a previous comment as well, in C++ it is not necessary to put every method in a class unlike object oriented languages like C#, so it is not necessary to bundle methods.

The instrument of choice is namespaces here 😄 Since all of these methods are static.

adjusted in: 4b1e81056c491c23bfe21335b43c7f07c3ffbb40

src/ObjLoading/Game/T6/BSP/BSPUtil.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include "BSP.h"
+
+namespace BSP
+{
+    class BSPUtil
+    {
+    public:
+        static std::string getFileNameForBSPAsset(std::string& assetName);
```
Laupetin: nit: Methods are generally named PascalCase in the codebase

adjusted in: 3d1fa7c6d9d80366143938b0f1910f71033066c2

Laupetin (commented):
I took a look at the entire code again this time and added a few comments.

As I mentioned before, **a lot** is just nitpicks and probably personal taste and since I was already on the branch, I just adjusted them quickly.
I added commits, split for each comment so any of them can be dropped again upon request.

If you don't like some of the changes I made, please tell me in the discussion 😅

asajid2-cell:
Hey! I’ve been testing the T6 custom-map branch against Plutonium and got a generated zm_cosmodrome map to reach live Zombies gameplay/HUD. I'm still working on understanding the engine; however, currently, I've found two small upstreamable issues:

1. FBX CullingOff was not honored for generated GfxWorld surfaces.
2. T6 generated world surfaces can silently accept model-material families like mc/*, but runtime rendering only became correct when using a wpc/* world material.

I have focused tests and runtime proof screenshots. Is there any way to contribute?

LJW-Dev:
Right now i'm moving to using GLTF as i found it to not have these issues that FBX has had, hopefully switching over would fix the 1st issue (the branch is updated to use gltf   ##now).

I agree a warning would be good when it sees not a world material, but when custom materials become possible i doubt many map makers would use the same wpc/wc/etc naming scheme so a full error would be too much. Maybe there's a way to tell if the material will/won't render apart from their naming scheme? 

Feel free to add a pull request to the branch if you want to add changes though.

asajid2-cell:
Sounds good, I’ll keep any future PR small and re-check against the current GLTF path first.

For context, my local fork/runtime lane can get a generated zm_cosmodrome into Plutonium ZM gameplay, including player spawn/movement and basic zombie melee damage, but it is still very early and has local probe/staging scaffolding. It was also based on your earlier custom-map work, so I’m now trying to line up with the current branch instead of guessing.

I tried a quick sanity check on the current bsp-compilation head with a minimal GLTF test map, but hit an early runtime crash before InitGame, so I may be misunderstanding the expected source setup.

Is there a specific current commit/example GLTF you consider the known-good in-game smoke test? I can base any testing/PR work on that.

Issue #298, opened 2024-11-02
Add Support to Fastfile for PS3 and Xbox 360


Darktreivor:
Hello, how are you? Thank you for the project!  
I would like to translate the games in the franchise for PS3 and Xbox 360 into Brazilian Portuguese, since only BO2 and later games originally came with a translation. Would it be possible to add support for PS3 and Xbox 360 fastfile, or guide me on how I can extract the ff files from these consoles to access the .str files for this purpose?  
I'll leave a sample of the code_post where the subtitles are supposedly located.

[code_post_gfx.zip](https://github.com/user-attachments/files/17607658/code_post_gfx.zip)

JezuzLizard:
Adding support for console versions is not trivial as OAT is not currently endian aware.
Also console structs may be different than PC structs which means there will need to be zone code generation per game, per platform.

Darktreivor:
> Adding support for console versions is not trivial as OAT is not currently endian aware. Also console structs may be different than PC structs which means there will need to be zone code generation per game, per platform.

I understand, but would it be possible to help me understand the file based on your experience with PC so that I can at least create an extractor and inserter for the .str?
My focus isn't on map or graphics mods, but solely on translating the game into other languages that lack support.

diamante0018:
No matter how small the change you want to make, everything Jezuz said needs to happen. You can't just say, let me insert a string into a fastfile because OAT does not even know how to open the file. A lot of work needs to be done to make OAT endian aware and then at the end, treat the console counter parts for each PC game like a completely separate game because the assets changed. Not to mention that to test anything to even confirm OAT works you need to rely of incomplete emulation tools or a modded console.

Laupetin:
Hey, thanks for the suggestion. I guess there is a bunch of reasons to support console fastfiles, it would be nice to be able to dump for example beta builds and also release builds.

The issue I see with what you are saying though is the following:
Idk how console fastfiles work in detail, but for PC the MP fastfiles are cryptographically signed which means that the game will refuse to load any fastfile that is not signed by Infinity Ward.
This is why only custom clients for MP can load custom fastfiles.
For SP this is not a requirement, so even Steam SP can load custom fastfiles.

Idk if that also applies to consoles so whether they are also signed. If they are though, it will not be possible to load custom fastfiles without a modified game.

Other than that supporting console fastfiles requires a lot of work to be done as the others mentioned above.
* It requires a solution for supporting big endian fastfiles and assets.
[...]

PR #611, opened 2025-12-14
feat: dump iw3 xbox unsigned fastfile data

michaeloliverx:
Adds support for dumping **unsigned** IW3 Xenon raw zones

Signed zone support is not included in this merge request, as it appears significantly more complex. This change is intended as an initial contribution to become familiar with the codebase.

Relates to https://github.com/Laupetin/OpenAssetTools/issues/298.

image


Laupetin (approved):
Seems to work great, thank you for the PR :)

Issue #546, opened 2025-10-23
support for exporting and importing glyph coordinate tables for fonts


3dll:
Hello, please tell me if there are plans to fully support the export of font-related files?
I want to do a fan localization, and I really need this feature. The atlas itself is stored as a texture in iwd, but in order to redraw it, I need the coordinates of the letters. I haven't been able to find this file anywhere, and as far as I understand, the program does not currently support its export and import. Is this planned for the future? 
Or maybe I'm doing something wrong)) If so, please help me with this)) Thank you)

Laupetin:
Implemented in #850 and #866

PR #739, opened 2026-04-13
Added Font_s dumper and loader

MrIkso:
This PRs adds feature to dump Font_s  to json structure and load it for create resource. Related to #546 
Supported games:

- [x] [IW3 (COD4: Modern Warfare - 2007)](https://en.wikipedia.org/wiki/Call_of_Duty_4:_Modern_Warfare)
- [x] [IW4 (Modern Warfare 2 - 2009)](https://en.wikipedia.org/wiki/Call_of_Duty:_Modern_Warfare_2)
- [x] [IW5 (Modern Warfare 3 - 2011)](https://en.wikipedia.org/wiki/Call_of_Duty:_Modern_Warfare_3)
- [x] [T5 (Black Ops - 2010)](https://en.wikipedia.org/wiki/Call_of_Duty:_Black_Ops)
- [ ] [T6 (Black Ops II - 2012)](https://en.wikipedia.org/wiki/Call_of_Duty:_Black_Ops_II)


src/ObjLoading/Game/IW3/Font/AssetLoaderFontIW3.cpp
```diff
@@ -0,0 +1,134 @@
+#include "AssetLoaderFontIW3.h"
+#include "Game/IW3/IW3.h"
```
Laupetin, line 2: The project uses clang-format for formatting. You should be able to configure clang-format as formatter in most IDEs.

src/ObjLoading/Game/IW3/Font/AssetLoaderFontIW3.cpp
```diff
@@ -0,0 +1,134 @@
...
+    public:
+        FontLoader(MemoryManager& memory, ISearchPath& searchPath)
+            : m_memory(memory),
+              m_search_path(searchPath)
+        {
+        }
+
+        AssetCreationResult CreateAsset(const std::string& assetName, AssetCreationContext& context) override
+        {
+            const auto file = m_search_path.Open(std::format("{}.json", assetName));
```
Laupetin, line 24: What other assets do is having a "common" file in the "ObjCommon" component that returns the file name for an asset, e.g. [MaterialCommon.h](https://github.com/Laupetin/OpenAssetTools/blob/ec8759494bd1f3bfbb37dd56f5c7f71af613e624/src/ObjCommon/Material/MaterialCommon.h).

This makes it possible to have a single source of truth for the filename pattern that is shared between dumpers and loaders.

src/ObjLoading/Game/IW3/Font/AssetLoaderFontIW3.cpp
```diff
@@ -0,0 +1,134 @@
...
+                jRoot.at("_type").get_to(type);
+
+                if (type != "font")
+                {
+                    con::error("Tried to load font but did not find expected type 'font'");
+                    return false;
+                }
+
+                font.pixelHeight = jRoot.value("pixelHeight", 0);
+                font.glyphCount = jRoot.value("glyphCount", 0);
```
Laupetin, line 59: This project uses the [arbitrary type conversions feature of nlohmann json](https://json.nlohmann.me/features/arbitrary_types), e.g. see [JsonLeaderboardDef.h](https://github.com/Laupetin/OpenAssetTools/blob/main/src/ObjCommon/Game/T6/Leaderboard/JsonLeaderboardDef.h) for the types and [JsonLoaderFontIconT6.cpp](https://github.com/Laupetin/OpenAssetTools/blob/ec8759494bd1f3bfbb37dd56f5c7f71af613e624/src/ObjLoading/Game/T6/FontIcon/JsonLoaderFontIconT6.cpp#L53) on them getting used.

Makes the loading code less verbose, prone to errors and makes the target format clearer in code imo.

Would be nice to not have this be an outlier in what style it uses😅

src/ObjLoading/Game/IW3/Font/AssetLoaderFontIW3.cpp
```diff
@@ -0,0 +1,134 @@
...
+                        const auto& jG = jGlyphs[i];
+                        auto& g = font.glyphs[i];
+
+                        g.letter = jG.value("letter", 0);
+
+                        g.x0 = static_cast<char>(jG.value("x0", 0));
+                        g.y0 = static_cast<char>(jG.value("y0", 0));
+                        g.dx = static_cast<char>(jG.value("dx", 0));
+                        g.pixelWidth = static_cast<char>(jG.value("width", 0));
+                        g.pixelHeight = static_cast<char>(jG.value("height", 0));
```
Laupetin, line 104: I made a mistake on the game types here, `dx`, `width` and `height` should be `unsigned char` (applies to all games).

src/ObjLoading/Game/IW3/Font/AssetLoaderFontIW3.cpp
```diff
@@ -0,0 +1,134 @@
...
+                if (type != "font")
+                {
+                    con::error("Tried to load font but did not find expected type 'font'");
+                    return false;
+                }
+
+                font.pixelHeight = jRoot.value("pixelHeight", 0);
+                font.glyphCount = jRoot.value("glyphCount", 0);
+
+                std::string fontName = jRoot.value("name", "");
```
Laupetin, line 61: The name part of the font asset should be taken from the input arguments of the loader (and the "name" should not be part of the json).

Otherwise you could try to load a font asset with one name and get an entirely different name (potentially by accident because you renamed the file but not the value in it)

Laupetin (commented):
Hey there, thanks for the PR :)

the json structure looks fine, i just have a few comments regarding the writing and loading methods to make it fit to the rest of the codebase.

I think the asset would be a good fit for having the code templated for between the games (like e.g. the [material asset](https://github.com/Laupetin/OpenAssetTools/blob/main/src/ObjLoading/Material/JsonMaterialLoader.cpp.template)) but i'm fine if you don't want to do this here as the templating format is a bit rough and odd 😅

src/ObjLoading/Game/IW3/Font/AssetLoaderFontIW3.cpp
MrIkso: okay i'm try add it formatter in visual studio, vs always used own standard 🤣

src/ObjLoading/Game/IW3/Font/AssetLoaderFontIW3.cpp
Laupetin: I like your usage of ordered_json and i noticed that the extensions for arbitrary type conversions for nlohmann json i defined only support the normal json and not ordered_json. I got to fix that i guess 😅

Laupetin: Did this here: #746 if you wanna use ordered_json still with arbitrary type conversions, feel free to rebase on main when its merged

PR #744, opened 2026-04-19
Add IW3 PhysPreset dump logic

pagingred:
This PR has logic to dump IW3 PhysPreset assets. I have tested this on 'physic/default' in the cod4 dev tools, and it dumps 1:1.

Only adds support for IW3


Laupetin:
hey there, thanks for the PR :) 
I don't know if you have seen it, but i've done a dumper (and loader) for PhysPreset for T6 (Black Ops 2).

The code there is oriented at what leftover code of the game can be found, so what the original game did.
See [PhysPresetInfoStringDumperT6.cpp](https://github.com/Laupetin/OpenAssetTools/blob/80dc487d5aaa36497645a986ca671ec27bcfebca/src/ObjWriting/Game/T6/PhysPreset/PhysPresetInfoStringDumperT6.cpp) (working on the field definitions in [PhysPresetFields.h](https://github.com/Laupetin/OpenAssetTools/blob/main/src/ObjCommon/Game/T6/PhysPreset/PhysPresetFields.h))

The IW3 struct should probably be very similar (maybe include a few fields less)? Would be nice if the code for IW3 would work similar to that 😅

src/ObjWriting/PhysPreset/PhysPresetDumpingZoneState.h
```diff
@@ -0,0 +1,23 @@
...
+    public:
+        class PhysPresetDumpingState
+        {
+        public:
+            std::string m_path;
+
+            PhysPresetDumpingState(std::string path);
+        };
+
+        std::map<const void*, PhysPresetDumpingState> m_physpreset_dumping_state_map;
```
Laupetin: What is this being used for? There only seems to be queries on it, the method that writes something in there `CreatePhysPresetDumpingState` is not being used?

src/ObjWriting/PhysPreset/PhysPresetDumpingZoneState.h
```diff
@@ -0,0 +1,23 @@
...
+
+        void CreatePhysPresetDumpingState(const void* menuDef, std::string path);
```
Laupetin: It is creating a state? I thought the class itself is already the state? Not sure if i understand the method name? 😅 Also the first parameter is called "menuDef"?

src/ObjWriting/PhysPreset/AbstractPhysPresetWriter.h
```diff
@@ -0,0 +1,38 @@
...
+#include "IPhysPresetWriter.h"
+
+#include <cstddef>
+#include <ostream>
+#include <string>
+#include <vector>
+
+namespace physpreset
+{
+    class AbstractBaseWriter : public IWriter
```
Laupetin: This is a very generic base class but it is contained inside a `physpreset` namespace? Judging by the methods there should not be anything physpreset specific in this, right?

src/ObjWriting/PhysPreset/AbstractPhysPresetWriter.h
```diff
@@ -0,0 +1,38 @@
...
+
+        void WriteRawU64(uint64_t val);
+        void WriteRawU32(uint32_t val);
+        void WriteRawU16(uint16_t val);
+        void WriteRawU8(uint8_t val);
+
+        void WriteRawI64(uint64_t val);
+        void WriteRawI32(uint32_t val);
+        void WriteRawI16(uint16_t val);
+        void WriteRawI8(uint8_t val);
```
Laupetin: None of these methods are used?

Laupetin (commented):
The code is a bit confusing to be honest. It looks like it went through a few iterations in the past that left a few remains 😅 
Maybe it can help to look a bit at the other dumpers I did (for other games) to get an idea on the style they usually follow in the codebase?

Not that that's the only way to write them, it's just nice to have a consistent style in a codebase.

pagingred:
Hey @Laupetin, thank you for the replies and review. I had used the IW4 menu dumping logic as a base to get this stuff done. I can see now that it was pretty overengineered and complicated for this, so I will go back and reference the dumper you had mentioned and get this cleaned up.

src/ObjWriting/PhysPreset/AbstractPhysPresetWriter.h
pagingred: I don't believe there would be anything in there tied directly to the physpreset asset. This can likely be moved out of the namespace or not used

src/ObjWriting/PhysPreset/PhysPresetDumpingZoneState.h
pagingred: Yeah I carried this stuff over from a previous implementation. Likely to be scrapped for the next review. 

The 'menuDef' parameter was also left over from the original implementation.

src/ObjWriting/PhysPreset/PhysPresetDumpingZoneState.h
pagingred: This will be removed for the next review, it is over-complicated and the logic is not required to dump such a simple asset.

pagingred (commented):
@Laupetin Now is much simpler and resembles the logic already present in T6.

src/ObjWriting/PhysPreset/AbstractPhysPresetWriter.h
pagingred: I simplified things a lot and do not use or need these abstract writers anymore.

src/ObjWriting/PhysPreset/PhysPresetDumpingZoneState.h
pagingred: Got rid of the whole dump state stuff, way too complicated.

src/ObjWriting/PhysPreset/PhysPresetDumpingZoneState.h
pagingred: Got rid of the whole dump state stuff, way too complicated.

src/ObjWriting/PhysPreset/AbstractPhysPresetWriter.h
pagingred: Switched over to the T6-style of dumping PhysPreset, so this whole abstract paradigm is gone now.

src/Common/Game/IW3/IW3_Assets.h
```diff
@@ -165,6 +165,21 @@ namespace IW3
...
+        float bounce;
+        float friction;
+        int isFrictionInfinity;
+        float bulletForceScale;
+        float explosiveForceScale;
+        const char* sndAliasPrefix;
+        float piecesSpreadFraction;
+        float piecesUpwardVelocity;
+        int tempDefaultToCylinder;
+        int perSurfaceSndAlias;
```
Laupetin: This field is not used, you probably copied it from IW4, but IW3 does not seem to have it.

src/ObjCommon/Game/IW3/PhysPreset/PhysPresetFields.h
```diff
@@ -0,0 +1,19 @@
...
+        {"sndAliasPrefix",        offsetof(PhysPresetInfo, sndAliasPrefix),        CSPFT_STRING  },
+        {"mass",                  offsetof(PhysPresetInfo, mass),                  CSPFT_FLOAT   },
+        {"friction",              offsetof(PhysPresetInfo, friction),              CSPFT_FLOAT   },
+        {"isFrictionInfinity",    offsetof(PhysPresetInfo, isFrictionInfinity),    CSPFT_QBOOLEAN},
+        {"bounce",                offsetof(PhysPresetInfo, bounce),                CSPFT_FLOAT   },
+        {"bulletForceScale",      offsetof(PhysPresetInfo, bulletForceScale),      CSPFT_FLOAT   },
+        {"explosiveForceScale",   offsetof(PhysPresetInfo, explosiveForceScale),   CSPFT_FLOAT   },
+        {"piecesSpreadFraction",  offsetof(PhysPresetInfo, piecesSpreadFraction),  CSPFT_FLOAT   },
+        {"piecesUpwardVelocity",  offsetof(PhysPresetInfo, piecesUpwardVelocity),  CSPFT_FLOAT   },
+        {"tempDefaultToCylinder", offsetof(PhysPresetInfo, tempDefaultToCylinder), CSPFT_FLOAT   },
```
Laupetin: This is not a float, this is CSPFT_QBOOLEAN

src/ObjCommon/Game/IW3/PhysPreset/PhysPresetFields.h
```diff
@@ -0,0 +1,19 @@
...
+    };
```
Laupetin, line 18: Can you bring these fields in the same order as they are in the struct? Makes it a bit easier to go through them.

src/ObjCommon/Game/IW3/ObjConstantsIW3.h
```diff
@@ -0,0 +1,21 @@
...
+        static constexpr const char* INFO_STRING_PREFIX_PHYS_PRESET = "PHYSIC";
+        static constexpr const char* INFO_STRING_PREFIX_TRACER = "TRACER";
+        static constexpr const char* INFO_STRING_PREFIX_VEHICLE = "VEHICLEFILE";
+        static constexpr const char* INFO_STRING_PREFIX_WEAPON = "WEAPONFILE";
+
+        static constexpr const char* GDF_FILENAME_MATERIAL = "material.gdf";
+        static constexpr const char* GDF_FILENAME_PHYS_PRESET = "physpreset.gdf";
+        static constexpr const char* GDF_FILENAME_TRACER = "tracer.gdf";
+        static constexpr const char* GDF_FILENAME_VEHICLE = "vehicle.gdf";
+        static constexpr const char* GDF_FILENAME_WEAPON = "weapon.gdf";
```
Laupetin: These constants are copied from IW4 i guess but they don't all make sense:
- IW3 does not have a tracer or vehicle asset
- IW4 only included material.gdf for some material decompilation experiments that won't make it to IW3 in this form, so that can be removed

src/ObjWriting/Game/IW3/PhysPreset/PhysPresetInfoStringDumperIW3.cpp
```diff
@@ -0,0 +1,105 @@
...
+            physPresetInfo->friction = physPreset->friction;
+        }
+
+        physPresetInfo->bounce = physPreset->bounce;
+        physPresetInfo->bulletForceScale = physPreset->bulletForceScale;
+        physPresetInfo->explosiveForceScale = physPreset->explosiveForceScale;
+        physPresetInfo->piecesSpreadFraction = physPreset->piecesSpreadFraction;
+        physPresetInfo->piecesUpwardVelocity = physPreset->piecesUpwardVelocity;
+        physPresetInfo->tempDefaultToCylinder = physPreset->tempDefaultToCylinder;
+    }
```
Laupetin: Please convert these fields in order as well to make it easier to compare to the struct (as was done in IW4)

src/ObjWriting/Game/IW3/PhysPreset/PhysPresetInfoStringDumperIW3.cpp
```diff
@@ -0,0 +1,105 @@
...
+        {
+        }
+    };
+
+    void CopyToPhysPresetInfo(const PhysPreset* physPreset, PhysPresetInfo* physPresetInfo)
+    {
+        physPresetInfo->sndAliasPrefix = physPreset->sndAliasPrefix;
+        physPresetInfo->mass = physPreset->mass;
+
+        if (physPreset->friction >= 3.40282e+38)
```
Laupetin:
```suggestion
        if (std::isinf(physPreset->friction))
```

This makes the intention more explicit without magic values.

src/ObjWriting/Game/IW3/PhysPreset/PhysPresetInfoStringDumperIW3.cpp
```diff
@@ -0,0 +1,105 @@
...
+            physPresetInfo->isFrictionInfinity = 0;
+            physPresetInfo->friction = physPreset->friction;
+        }
+
+        physPresetInfo->bounce = physPreset->bounce;
+        physPresetInfo->bulletForceScale = physPreset->bulletForceScale;
+        physPresetInfo->explosiveForceScale = physPreset->explosiveForceScale;
+        physPresetInfo->piecesSpreadFraction = physPreset->piecesSpreadFraction;
+        physPresetInfo->piecesUpwardVelocity = physPreset->piecesUpwardVelocity;
+        physPresetInfo->tempDefaultToCylinder = physPreset->tempDefaultToCylinder;
```
Laupetin:
```suggestion
        physPresetInfo->tempDefaultToCylinder = physPreset->tempDefaultToCylinder ? 1 : 0;
```

the `PhysPreset` struct has a `bool` -> 1 byte and the info struct has an `int` -> 4 byte.
So please make the values explicit please 😅 

IW3's original game code not doing this is a bug.

src/Common/Game/IW3/IW3_Assets.h
pagingred: I sure did, this extra field is now removed.

src/ObjCommon/Game/IW3/PhysPreset/PhysPresetFields.h
pagingred: Changed to CSPFT_QBOOLEAN.

src/ObjCommon/Game/IW3/PhysPreset/PhysPresetFields.h
pagingred: I originally changed these so the dumped file would match the original, but since order doesn't seem to matter for these I get where you're coming from. Changed to match the struct.

src/ObjCommon/Game/IW3/ObjConstantsIW3.h
pagingred: Tracer defs, vehicle defs, and material def removed.

src/ObjWriting/Game/IW3/PhysPreset/PhysPresetInfoStringDumperIW3.cpp
pagingred: Changed to match the struct.

src/ObjWriting/Game/IW3/PhysPreset/PhysPresetInfoStringDumperIW3.cpp
pagingred: Completely missed this when templating from the original. Changed to clamp mass and tempDefaultToCylinder.

NOTE: The original clamp statement used for mass was not functioning properly for IW3. It pretty much always returned the max since I was multiplying the mass by 1000. Just removing this multiplication fixes it so I assume it is a kg vs g change across the engines.

src/ObjWriting/Game/IW3/PhysPreset/PhysPresetInfoStringDumperIW3.cpp
pagingred: I completely agree and was hoping to use the isinf function. However, the reason I added the magic value is because isinf was not indicating that the friction was at infinity. Instead isFrictionInfinity was equal to 0 and friction was equal to 3.40282e+38. Let me know how you want to proceed.

pagingred:
@Laupetin I implemented those changes you suggested. Thank you for being patient with this PR, I hope future ones will be smoother.

src/ObjWriting/Game/IW3/PhysPreset/PhysPresetInfoStringDumperIW3.cpp
Laupetin: Hm okay, that's true, IW3 seems to use different logic here.
What works though and is a lot nicer is:
```suggestion
        if (physPreset->friction >= std::numeric_limits<float>::max())
```

You will need to `#include <limits>`. When you do that you can also get rid of the unnecessary "cmath" include as well :)

src/ObjWriting/Game/IW3/PhysPreset/PhysPresetInfoStringDumperIW3.cpp
Laupetin: The code from IW4 was based on what Black Ops 1 does.
Not every game includes code that actually parses these files so for some titles it's guess work on what logic was behind these values.

Black Ops 1 _does_ (when loading) do clamping between 1.0 and 2000.0 and then multiply with 0.001

However IW3 is in fact a game that includes the original load code as well and the logic there does neither the clamping nor the multiplication.

So yeah you are right in that the multiplication is wrong here. But the clamping is as well 😅 Especially since it works on the 1000 times higher values. So you are clamping most values to 1 here even though it shouldn't.

For Cod4: Not sure if that game _actually_ does this multiplication and clamping.
If it did, there should be no values from fastfiles that would actually be clamped.
Idk if i ever checked that.
The chances are good that game also does not clamp and multiply as Cod4 was the game after which the IW games and Treyarch games codebases started to diverge a bunch.
But that's for another time.

Laupetin: btw: while the IW3 code that loads these files does not clamp the values, the official mod tools asset manager set a limit like this: `float 1.0 0.01 10000.0 mass`, so a min of 0.01 and a max of 10000.0.
That's just the UI though.

Laupetin:
> @Laupetin I implemented those changes you suggested. Thank you for being patient with this PR, I hope future ones will be smoother.

No worries 😄 Code review comments are normal and don't mean a PR is unsmooth.

src/ObjWriting/Game/IW3/PhysPreset/PhysPresetInfoStringDumperIW3.cpp
Laupetin: testing iw4 physpreset values and it does actually clamp for a bunch of assets. So it most likely uses the cod4 logic then. Guess that should be fixed then 😬

src/ObjWriting/Game/IW3/PhysPreset/PhysPresetInfoStringDumperIW3.cpp
pagingred: Thanks! Nice solution, I just added it in.

src/ObjWriting/Game/IW3/PhysPreset/PhysPresetInfoStringDumperIW3.cpp
pagingred: I gotcha, things get sticky when thinking about differences across engines lol. I removed the clamping entirely for the time being.

Laupetin (approved):
Thanks for the PR :) added some cleanup refactoring (of my past code) for ObjConstants at the end as I thought it was needed and the place was touched here anyway

pagingred:
I started using the clang-format scripts so going forward I won't waste as many CI/CD runs. Thank you for your help on the PR!

PR #753, opened 2026-04-23
feat: phys presets dumping and loading for IW3, IW5, T5

pagingred:
This PR has logic to dump and load PhysPreset assets for IW3, IW5, T5, and T6.

### DIFF
--- a/src/Common/Game/IW5/IW5.h
+++ b/src/Common/Game/IW5/IW5.h
@@ -197,6 +197,13 @@ namespace IW5
         WAFT_NUM_FIELD_TYPES,
     };
 
+    enum physPresetFieldType_t
+    {
+        PPFT_SCALING = WAFT_NUM_FIELD_TYPES,
+
+        PPFT_NUM_FIELD_TYPES,
+    };
+
     using AssetPhysPreset = Asset<ASSET_TYPE_PHYSPRESET, PhysPreset>;
     using AssetPhysCollMap = Asset<ASSET_TYPE_PHYSCOLLMAP, PhysCollmap>;
     using AssetXAnim = Asset<ASSET_TYPE_XANIMPARTS, XAnimParts>;
--- a/src/Common/Game/IW5/IW5.h
+++ b/src/Common/Game/IW5/IW5.h
@@ -199,7 +199,7 @@ namespace IW5
 
     enum physPresetFieldType_t
     {
-        PPFT_SCALING = WAFT_NUM_FIELD_TYPES,
+        PPFT_SCALING = CSPFT_NUM_BASE_FIELD_TYPES,
 
         PPFT_NUM_FIELD_TYPES,
     };
--- a/src/Common/Game/IW5/IW5_Assets.h
+++ b/src/Common/Game/IW5/IW5_Assets.h
@@ -203,8 +203,6 @@ namespace IW5
 
     struct PhysPresetInfo
     {
-        const char* name;
-        int type;
         float mass;
         float bounce;
         float friction;
--- a/src/Common/Game/IW5/IW5_Assets.h
+++ b/src/Common/Game/IW5/IW5_Assets.h
@@ -201,6 +201,28 @@ namespace IW5
         bool perSurfaceSndAlias;
     };
 
+    struct PhysPresetInfo
+    {
+        const char* name;
+        int type;
+        float mass;
+        float bounce;
+        float friction;
+        float bulletForceScale;
+        float explosiveForceScale;
+        const char* sndAliasPrefix;
+        float piecesSpreadFraction;
+        float piecesUpwardVelocity;
+        float minMomentum;
+        float maxMomentum;
+        float minPitch;
+        float maxPitch;
+        PhysPresetScaling volumeType;
+        PhysPresetScaling pitchType;
+        int tempDefaultToCylinder;
+        int perSurfaceSndAlias;
+    };
+
     struct Bounds
     {
         vec3_t midPoint;
--- a/src/Common/Game/T5/T5_Assets.h
+++ b/src/Common/Game/T5/T5_Assets.h
@@ -185,7 +185,6 @@ namespace T5
         int isFrictionInfinity;
         float bulletForceScale;
         float explosiveForceScale;
-        const char* sndAliasPrefix;
         float piecesSpreadFraction;
         float piecesUpwardVelocity;
         int canFloat;
--- a/src/Common/Game/T5/T5_Assets.h
+++ b/src/Common/Game/T5/T5_Assets.h
@@ -177,6 +177,24 @@ namespace T5
         vec3_t buoyancyBoxMax;
     };
 
+    struct PhysPresetInfo
+    {
+        float mass;
+        float bounce;
+        float friction;
+        int isFrictionInfinity;
+        float bulletForceScale;
+        float explosiveForceScale;
+        const char* sndAliasPrefix;
+        float piecesSpreadFraction;
+        float piecesUpwardVelocity;
+        int canFloat;
+        float gravityScale;
+        vec3_t centerOfMassOffset;
+        vec3_t buoyancyBoxMin;
+        vec3_t buoyancyBoxMax;
+    };
+
     enum ConstraintType
     {
         CONSTRAINT_NONE = 0x0,
--- a/src/ObjCommon/Game/IW5/PhysPreset/PhysPresetFields.h
+++ b/src/ObjCommon/Game/IW5/PhysPreset/PhysPresetFields.h
@@ -5,8 +5,6 @@
 namespace IW5
 {
     inline cspField_t phys_preset_fields[]{
-        {"name",                  offsetof(PhysPresetInfo, name),                  CSPFT_STRING  },
-        {"type",                  offsetof(PhysPresetInfo, type),                  CSPFT_INT     },
         {"mass",                  offsetof(PhysPresetInfo, mass),                  CSPFT_FLOAT   },
         {"bounce",                offsetof(PhysPresetInfo, bounce),                CSPFT_FLOAT   },
         {"friction",              offsetof(PhysPresetInfo, friction),              CSPFT_FLOAT   },
--- a/src/ObjCommon/Game/IW5/PhysPreset/PhysPresetFields.h
+++ b/src/ObjCommon/Game/IW5/PhysPreset/PhysPresetFields.h
@@ -4,11 +4,6 @@
 
 namespace IW5
 {
-    inline const char* szPhysPresetScalingNames[]{
-        "linear",
-        "quadratic",
-    };
-
     inline cspField_t phys_preset_fields[]{
         {"name",                  offsetof(PhysPresetInfo, name),                  CSPFT_STRING  },
         {"type",                  offsetof(PhysPresetInfo, type),                  CSPFT_INT     },
--- a/src/ObjCommon/Game/IW5/PhysPreset/PhysPresetFields.h
+++ b/src/ObjCommon/Game/IW5/PhysPreset/PhysPresetFields.h
@@ -25,5 +25,3 @@ namespace IW5
         {"perSurfaceSndAlias",    offsetof(PhysPresetInfo, perSurfaceSndAlias),    CSPFT_QBOOLEAN},
     };
 }
-
-
--- a/src/ObjCommon/Game/IW5/PhysPreset/PhysPresetFields.h
+++ b/src/ObjCommon/Game/IW5/PhysPreset/PhysPresetFields.h
@@ -1,0 +1,34 @@
+#pragma once
+
+#include "Game/IW5/IW5.h"
+
+namespace IW5
+{
+    inline const char* szPhysPresetScalingNames[]{
+        "linear",
+        "quadratic",
+    };
+
+    inline cspField_t phys_preset_fields[]{
+        {"name",                  offsetof(PhysPresetInfo, name),                  CSPFT_STRING  },
+        {"type",                  offsetof(PhysPresetInfo, type),                  CSPFT_INT     },
+        {"mass",                  offsetof(PhysPresetInfo, mass),                  CSPFT_FLOAT   },
+        {"bounce",                offsetof(PhysPresetInfo, bounce),                CSPFT_FLOAT   },
+        {"friction",              offsetof(PhysPresetInfo, friction),              CSPFT_FLOAT   },
+        {"bulletForceScale",      offsetof(PhysPresetInfo, bulletForceScale),      CSPFT_FLOAT   },
+        {"explosiveForceScale",   offsetof(PhysPresetInfo, explosiveForceScale),   CSPFT_FLOAT   },
+        {"sndAliasPrefix",        offsetof(PhysPresetInfo, sndAliasPrefix),        CSPFT_STRING  },
+        {"piecesSpreadFraction",  offsetof(PhysPresetInfo, piecesSpreadFraction),  CSPFT_FLOAT   },
+        {"piecesUpwardVelocity",  offsetof(PhysPresetInfo, piecesUpwardVelocity),  CSPFT_FLOAT   },
+        {"minMomentum",           offsetof(PhysPresetInfo, minMomentum),           CSPFT_FLOAT   },
+        {"maxMomentum",           offsetof(PhysPresetInfo, maxMomentum),           CSPFT_FLOAT   },
+        {"minPitch",              offsetof(PhysPresetInfo, minPitch),              CSPFT_FLOAT   },
+        {"maxPitch",              offsetof(PhysPresetInfo, maxPitch),              CSPFT_FLOAT   },
+        {"volumeType",            offsetof(PhysPresetInfo, volumeType),            PPFT_SCALING  },
+        {"pitchType",             offsetof(PhysPresetInfo, pitchType),             PPFT_SCALING  },
+        {"tempDefaultToCylinder", offsetof(PhysPresetInfo, tempDefaultToCylinder), CSPFT_QBOOLEAN},
+        {"perSurfaceSndAlias",    offsetof(PhysPresetInfo, perSurfaceSndAlias),    CSPFT_QBOOLEAN},
+    };
+}
+
+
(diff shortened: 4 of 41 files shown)


pagingred:
Points of Clarification:

- Currently the PhysPresetInfo struct is the same as the PhysPreset struct. Should I just use the PhysPreset one instead? Or should the info structure be different and I left something out?

- I was unable to find original source physic assets to verify against, so I had to make assumptions. The main one being that the volumeType and pitchType fields get converted to a string value. This is how I saw similar enums being handled within the WeaponDef, so I figured it was done in a similar way here.

pagingred:
Not sure why these checks failed, this commit hash builds fine on my machine.

Laupetin:
> Not sure why these checks failed, this commit hash builds fine on my machine.

Try to rebase your branch on the current main branch. It is due to the changes I made on the last PR, I refactored the ObjConstants files 😅

The CI does not run on the last commit of your branch but on a hypothetical merge commit after it would have been merged in main. So it considers the latest changes from main in it it's result.

Laupetin:
> Currently the PhysPresetInfo struct is the same as the PhysPreset struct. Should I just use the PhysPreset one instead? Or should the info structure be different and I left something out?

The info struct is already different in your case, `tempDefaultToCylinder` and `perSurfaceSndAlias` are `int` as they use QBOOLEAN.

Also the field `type` exists in every game and is consistently not dumped or parsed.
I assume the game uses it for runtime information?
Not exactly sure whats up with that, but the game games also omit that from their info struct.

src/Common/Game/IW5/IW5.h
```diff
@@ -197,6 +197,13 @@ namespace IW5
         WAFT_NUM_FIELD_TYPES,
     };
 
+    enum physPresetFieldType_t
+    {
+        PPFT_SCALING = WAFT_NUM_FIELD_TYPES,
```
Laupetin:
```suggestion
        PPFT_SCALING = CSPFT_NUM_BASE_FIELD_TYPES,
```

this continues the field types from attachment which is not tragic but also does not make sense and looks like a mistake.

Laupetin:
> I was unable to find original source physic assets to verify against, so I had to make assumptions. The main one being that the volumeType and pitchType fields get converted to a string value. This is how I saw similar enums being handled within the WeaponDef, so I figured it was done in a similar way here.

That happens a lot with IW4, IW5 unfortunately. Would have done the same thing here.
I just try to be consistent with other games if they do something similar.

If not, if there is no way to prove it wrong, there's no problem with it really, as it might just as well be the original way.
And if not, there's no original assets or code that would be incompatible.

src/Common/Game/IW5/IW5_Assets.h
```diff
@@ -201,6 +201,28 @@ namespace IW5
         bool perSurfaceSndAlias;
     };
 
+    struct PhysPresetInfo
+    {
+        const char* name;
+        int type;
```
Laupetin: These two should be omitted from the info

src/ObjCommon/Game/IW5/PhysPreset/PhysPresetFields.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include "Game/IW5/IW5.h"
+
+namespace IW5
+{
+    inline cspField_t phys_preset_fields[]{
+        {"name",                  offsetof(PhysPresetInfo, name),                  CSPFT_STRING  },
+        {"type",                  offsetof(PhysPresetInfo, type),                  CSPFT_INT     },
```
Laupetin: These two should not be dumped:

## Name
The `name` of the asset should be defined entirely by the filename. Otherwise two problems arise: 
1. If you want to rename or copy an asset, you need to remember to both rename and edit the name in the file, otherwise it is inconsistent 
2. you may load an asset with name X but an asset with name Y gets added and you wouldn't know why and the build would fail.

## Type

The type is not dumped by any game and is always left 0, compare with the other games including IW3

src/ObjWriting/Game/IW5/PhysPreset/PhysPresetInfoStringDumperIW5.cpp
```diff
@@ -0,0 +1,130 @@
...
+            : InfoStringFromStructConverter(structure, fields, fieldCount, std::move(scriptStringValueCallback))
+        {
+        }
+    };
+
+    void CopyToPhysPresetInfo(const PhysPreset* physPreset, PhysPresetInfo* physPresetInfo)
+    {
+        physPresetInfo->name = physPreset->name;
+        physPresetInfo->type = physPreset->type;
+        physPresetInfo->mass = std::clamp(physPreset->mass, 1.0f, 2000.0f);
```
Laupetin: Don't clamp here, I checked on IW4 and it seems to be incorrect there. IW5 is very likely the same..

src/ObjWriting/Game/IW5/PhysPreset/PhysPresetInfoStringDumperIW5.cpp
```diff
@@ -0,0 +1,130 @@
...
+        }
+        if (physPreset->pitchType == PHYSPRESET_SCALING_LINEAR || physPreset->pitchType == PHYSPRESET_SCALING_QUADRATIC)
+        {
+            physPresetInfo->pitchType = physPreset->pitchType;
+        }
+        else
+        {
+            // If type is invalid, assume linear
+            physPresetInfo->pitchType = PHYSPRESET_SCALING_LINEAR;
+        }
```
Laupetin: The game struct should not have any values that are not that.
Even when it does, the `FillFromEnumInt` has a fallback for it.
I'd leave that entire code away and just do:

```
physPresetInfo->volumeType = PHYSPRESET_SCALING_LINEAR;
physPresetInfo->pitchType = PHYSPRESET_SCALING_LINEAR;
```

pagingred:
> > Not sure why these checks failed, this commit hash builds fine on my machine.
> 
> Try to rebase your branch on the current main branch. It is due to the changes I made on the last PR, I refactored the ObjConstants files 😅
> 
> The CI does not run on the last commit of your branch but on a hypothetical merge commit after it would have been merged in main. So it considers the latest changes from main in it it's result.

So this is my first set of contributions via a fork and PRs. I now see that when the upstream is updated I need to sync my fork, so that is now done and after this push everything should build.

src/ObjWriting/Game/IW5/PhysPreset/PhysPresetInfoStringDumperIW5.cpp
pagingred: > The game struct should not have any values that are not that. Even when it does, the `FillFromEnumInt` has a fallback for it. I'd leave that entire code away and just do:
> 
> ```
> physPresetInfo->volumeType = PHYSPRESET_SCALING_LINEAR;
> physPresetInfo->pitchType = PHYSPRESET_SCALING_LINEAR;
> ```

Just for clarification, did you mean:

physPresetInfo->volumeType = physPreset->volumeType;
physPresetInfo->pitchType = physPreset->pitchType;

Since assigning them to LINEAR within CopyToPhysPresetInfo would not utilize the volumeType available on PhysPreset.

src/Common/Game/T5/T5_Assets.h
```diff
@@ -177,6 +177,24 @@ namespace T5
...
 
+    struct PhysPresetInfo
+    {
+        float mass;
+        float bounce;
+        float friction;
+        int isFrictionInfinity;
+        float bulletForceScale;
+        float explosiveForceScale;
+        const char* sndAliasPrefix;
```
Laupetin: T5 does not seem to use this field. This can be omitted from the PhysPresetInfo struct.

src/ObjCommon/Game/T5/PhysPreset/PhysPresetFields.h
```diff
@@ -0,0 +1,27 @@
...
+    inline cspField_t phys_preset_fields[]{
+        {"mass",                 offsetof(PhysPresetInfo, mass),                 CSPFT_FLOAT   },
+        {"bounce",               offsetof(PhysPresetInfo, bounce),               CSPFT_FLOAT   },
+        {"friction",             offsetof(PhysPresetInfo, friction),             CSPFT_FLOAT   },
+        {"isFrictionInfinity",   offsetof(PhysPresetInfo, isFrictionInfinity),   CSPFT_QBOOLEAN},
+        {"bulletForceScale",     offsetof(PhysPresetInfo, bulletForceScale),     CSPFT_FLOAT   },
+        {"explosiveForceScale",  offsetof(PhysPresetInfo, explosiveForceScale),  CSPFT_FLOAT   },
+        {"piecesSpreadFraction", offsetof(PhysPresetInfo, piecesSpreadFraction), CSPFT_FLOAT   },
+        {"piecesUpwardVelocity", offsetof(PhysPresetInfo, piecesUpwardVelocity), CSPFT_FLOAT   },
+        {"canFloat",             offsetof(PhysPresetInfo, canFloat),             CSPFT_INT     },
```
Laupetin: It is correct that the game does consider this `CSPFT_INT`, though this seems odd to me and enables inputs that are not intended.
The mod tools `gdf` defines this field as `int 1 0 1 canFloat` which means min 0, max 1.
I also checked against all physpreset assets of the vanilla game, there is none that has a value that is not `0` or `1`.

So i'd make this `CSPFT_QBOOLEAN`?

src/ObjLoading/Game/IW3/PhysPreset/InfoStringLoaderPhysPresetIW3.cpp
```diff
@@ -0,0 +1,90 @@
...
+        bool ConvertExtensionField(const cspField_t& field, const std::string& value) override
+        {
+            assert(false);
+            return false;
+        }
+    };
+
+    void CopyFromPhysPresetInfo(const PhysPresetInfo& physPresetInfo, PhysPreset& physPreset)
+    {
+        physPreset.mass = std::clamp(physPresetInfo.mass, 1.0f, 2000.0f) * 0.001f;
```
Laupetin: IW3 dumper does no clamping and multiplication

src/ObjLoading/Game/IW3/PhysPreset/InfoStringLoaderPhysPresetIW3.cpp
```diff
@@ -0,0 +1,90 @@
...
+        physPreset.bounce = physPresetInfo.bounce;
+
+        if (physPresetInfo.isFrictionInfinity != 0)
+            physPreset.friction = std::numeric_limits<float>::infinity();
```
Laupetin: As you correctly noticed when doing the dumper: IW3 does not use `std::numeric_limits<float>::infinity()` but `std::numeric_limits<float>::max()` here

src/ObjLoading/Game/IW5/PhysPreset/InfoStringLoaderPhysPresetIW5.cpp
```diff
@@ -0,0 +1,83 @@
...
+                                        size_t fieldCount)
+            : InfoStringToStructConverter(infoString, structure, zoneScriptStrings, memory, context, registration, fields, fieldCount)
+        {
+        }
+
+    protected:
+        bool ConvertExtensionField(const cspField_t& field, const std::string& value) override
+        {
+            assert(false);
+            return false;
```
Laupetin: This is missing an implementation for `PPFT_SCALING`, which is an extension field in IW5 that was added.

src/ObjLoading/Game/IW5/PhysPreset/InfoStringLoaderPhysPresetIW5.cpp
```diff
@@ -0,0 +1,83 @@
...
+        physPreset.mass = physPresetInfo.mass;
+        physPreset.bounce = physPresetInfo.bounce;
+        physPreset.friction = physPresetInfo.friction;
+        physPreset.bulletForceScale = physPresetInfo.bulletForceScale;
+        physPreset.explosiveForceScale = physPresetInfo.explosiveForceScale;
+        physPreset.sndAliasPrefix = physPresetInfo.sndAliasPrefix;
+        physPreset.piecesSpreadFraction = physPresetInfo.piecesSpreadFraction;
+        physPreset.piecesUpwardVelocity = physPresetInfo.piecesUpwardVelocity;
+        physPreset.tempDefaultToCylinder = physPresetInfo.tempDefaultToCylinder != 0;
+        physPreset.perSurfaceSndAlias = physPresetInfo.perSurfaceSndAlias != 0;
```
Laupetin, line 64: This is missing a bunch of IW5 fields, including `minMomentum`, `maxMomentum`, `minPitch`, `maxPitch`, `volumeType`, `pitchType`

src/ObjLoading/Game/T5/InfoString/InfoStringToStructConverter.cpp
```diff
@@ -0,0 +1,197 @@
...
+        if (!value.empty() && value[0] == '@')
+        {
+            char* endPtr;
+            hash = strtoul(&value[1], &endPtr, 16);
+            return endPtr == &value[value.size()];
+        }
+
+        hash = Common::Com_HashString(value.c_str());
+        return true;
+    }
```
Laupetin: You copied this from T6, it is not used here.

src/ObjLoading/Game/T5/PhysPreset/InfoStringLoaderPhysPresetT5.cpp
```diff
@@ -0,0 +1,100 @@
...
+        physPreset.bounce = physPresetInfo.bounce;
+
+        if (physPresetInfo.isFrictionInfinity != 0)
+            physPreset.friction = std::numeric_limits<float>::infinity();
+        else
+            physPreset.friction = physPresetInfo.friction;
+
+        physPreset.bulletForceScale = physPresetInfo.bulletForceScale;
+        physPreset.explosiveForceScale = physPresetInfo.explosiveForceScale;
+        physPreset.sndAliasPrefix = physPresetInfo.sndAliasPrefix;
```
Laupetin: `sndAliasPrefix` is not used in T5, can be omitted

src/ObjWriting/Game/T5/PhysPreset/PhysPresetInfoStringDumperT5.cpp
```diff
@@ -0,0 +1,108 @@
...
+        {
+        }
+    };
+
+    void CopyToPhysPresetInfo(const PhysPreset* physPreset, PhysPresetInfo* physPresetInfo)
+    {
+        physPresetInfo->mass = std::clamp(physPreset->mass * 1000.0f, 1.0f, 2000.0f);
+        physPresetInfo->bounce = physPreset->bounce;
+
+        if (std::isinf(physPreset->friction))
```
Laupetin: Treyarch apparently thought that std::numeric_limits<float>::max() is boring and set the limit to 1e+10 (not a technical constant i know of).
So i guess constexpr float MAX_FRICTION = 1e+10; it is here (within the anonymous namespace).
And then

```suggestion
        if (physPreset->friction >= MAX_FRICTION)
```

here.
Funnily the same thing goes for T6 as well, which is also wrong then.

src/ObjLoading/Game/T5/PhysPreset/InfoStringLoaderPhysPresetT5.cpp
```diff
@@ -0,0 +1,100 @@
...
+        }
+    };
+
+    void CopyFromPhysPresetInfo(const PhysPresetInfo& physPresetInfo, PhysPreset& physPreset)
+    {
+        physPreset.mass = std::clamp(physPresetInfo.mass, 1.0f, 2000.0f) * 0.001f;
+        physPreset.bounce = physPresetInfo.bounce;
+
+        if (physPresetInfo.isFrictionInfinity != 0)
+            physPreset.friction = std::numeric_limits<float>::infinity();
```
Laupetin: This needs the same change as the dumper then.

```suggestion
            physPreset.friction = MAX_FRICTION;
```

Laupetin (commented):
Thanks for the PR :) 
pretty nice to have the PhysPreset covered everywhere.
I noticed the IW5 physpresets did not load though, please make sure to always test the changes (even when you could reuse a lot of code that was previously there) 😅 

I did the follow to verify that all physpresets of the vanilla game load and build into identical data (i am using zsh, the exact commands are for that or bash, not for windows cmd or powershell; and also paths are from my machine):

```shell
# Dump all vanilla physpreset assets into zone_raw/iw5_physics
./Unlinker -v -o ./zone_raw/iw5_physics --include-assets "physpreset" C:/Cod/IW5/zone/english/*.ff

# Remove unnecessary zone files
rm zone_raw/iw5_physics/zone_source/*.zone

# Generate a .zone will all available physpresets
(echo ">game,t5\n" && ls -w 1 ./zone_raw/t5_physics/physic | sed 's/^/physpreset,/') > ./zone_raw/t5_physics/zone_source/t5_physics.zone

# Create fastfile with all physpreset assets
./Linker -v t5_physics

# Unlink that fastfile again
./Unlinker -v ./zone_out/t5_physics/t5_physics.ff

# Compare all files from output to input
# Differences are printed
# No output means no differences
diff -ur ./zone_raw/t5_physics/physic ./zone_dump/zone_raw/t5_physics/physic
```

that's only my solution though as an example, how that is done exactly is not exactly important 😛

src/ObjLoading/Game/IW5/PhysPreset/InfoStringLoaderPhysPresetIW5.cpp
```diff
@@ -0,0 +1,97 @@
...
+                                        const cspField_t* fields,
+                                        size_t fieldCount)
+            : InfoStringToStructConverter(infoString, structure, zoneScriptStrings, memory, context, registration, fields, fieldCount)
+        {
+        }
+
+    protected:
+        bool ConvertExtensionField(const cspField_t& field, const std::string& value) override
+        {
+            switch (static_cast<csParseFieldType_t>(field.iFieldType))
```
Laupetin: As clang warning says:
```suggestion
            switch (static_cast<physPresetFieldType_t>(field.iFieldType))
```

Laupetin (approved):
thank you for completing the PhysPreset asset :D

PR #760, opened 2026-04-27
fix: xmodel glTF dumps dropping `COLOR_0` vertex attributes

michaeloliverx:
I ran into this while working in my Xenon fork, but the root cause turned out to be upstream.

Preserve `COLOR_0` in xmodel glTF dumps. The writer was dropping vertex colors from dumped GLBs, causing readers to fall back to default white and breaking flag/cloth deformation masks.

The video shows the original map on the left and an OAT-linked version on the right using the `.glb` loader for `xmodel,bro_tarp_crate_anim` from `mp_osg_cargo`.

https://github.com/user-attachments/assets/f43638cc-64ea-4c5b-958e-5cbe92202fd6

EDIT: You can also observe the issue on stockmaps with flags e.g. `mp_showdown`

### DIFF
--- a/src/ObjCommon/XModel/Gltf/JsonGltf.h
+++ b/src/ObjCommon/XModel/Gltf/JsonGltf.h
@@ -265,7 +265,7 @@ namespace gltf
         std::optional<unsigned> WEIGHTS_0;
     };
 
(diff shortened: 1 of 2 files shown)


Laupetin (approved):
Good find, thanks :D I adjusted the code a bit to only dump the color data when there is non-default data, meaning not every vertex has a color value of 1,1,1,1.
Just so that it's maybe easier to work with models that do not require handling a separate color layer on vertices.

PR #761, opened 2026-04-27
fix: IW3 ensure `XModel::surfs` is set in loader

michaeloliverx:


Laupetin (approved):
Thanks :) Good find and also whoops

PR #763, opened 2026-04-28
Add IW3 PhysPreset Loading Logic

pagingred:
This PR adds in functionality to load in an IW3 PhysPreset either from the GDT or from the local disk. It was created as nearly a 1:1 copy of the IW4 loader with the only significant changes being:
- Adjusted to not dump 'perSurfaceSndAlias' (not in IW3)
- Kept 'isFrictionInfinity' check and logic. Even though IW3 struct doesn't seem to have this.


pagingred:

PR #764, opened 2026-04-28
Add IW5 PhysPreset Loading Logic

pagingred:
This PR adds in functionality to load in an IW5 PhysPreset either from the GDT or from the local disk. It was created as nearly a 1:1 copy of the IW4 loader with the only significant changes being:
- Added types and parsing for scaling types like volumeType and pitchType.
- Removed 'isFrictionInfinity' check and logic.


pagingred:

PR #765, opened 2026-04-29
Add T5 PhysPreset Dumping & Loading Logic

pagingred:
This PR adds in functionality to load a T5 PhysPreset either from the GDT or from the local disk and dump one. It was created as nearly a 1:1 copy of the T6 loader/dumper with the only significant changes being:
- Added dumping for sndAlias prefix. (Not having this seems inconsistent between T6 v.s. IW3, IW4, T5)


pagingred:

PR #766, opened 2026-04-29
Add IW3 & IW5 SndCurve Dumping & Loading Logic

pagingred:
This PR adds in functionality to load and dump an IW3/IW5 SndCurve. It was created as nearly a 1:1 copy of the IW4 loader/dumper.


pagingred:
Should we maintain a standard for this asset name? I have seen SoundCurve and SndCurve in files/classes and sound_curve and snd_curve as namespaces.

Laupetin:
> Should we maintain a standard for this asset name? I have seen SoundCurve and SndCurve in files/classes and sound_curve and snd_curve as namespaces.

I am usually not renaming game structs as it makes it harder to cross reference with IDA or other places.
But for other place i guess it should probably be "sound curve" instead of "snd curve".

So `AssetSoundCurve` and the namespace `sound_curve`

Laupetin (approved):
looks good, thanks

PR #767, opened 2026-04-29
Add snd_alias_list_t (IW3-5) Dumping/Loading Logic and LoadedSound (IW3-5) Loading Logic

pagingred:
WORK IN PROGRESS

This PR adds in functionality to load and dump a snd_alias_list_t asset for IW3, IW4, and IW5. As there was no previous functionality for a feature like this I chose to model it based off of the InfoString dumpers. I did this with the idea of creating a CSV dumper and loader that would work in a similar way.


pagingred:
There is still significant work to be done on this feature. AFAIK the process for compiling these alias lists is lossy. 
1. The min and max volume are multiplied by the set volume modifier group's value. At this point I cannot discern what the original group was per sound alias (working on a heuristic) so I cannot determine the correct original volume.
2. The volume modifier group itself, since it cannot be determined rn, is not dumped. Future games (like AW) have a separate field for this 'volModIndex' which makes me believe this value is present in the form of a bit masked value in the flags field.
3. Alias type (loaded, streamed, primed) is currently based on bits 6-7 of the flags. Might be wrong
4. Loop status (nonlooping, looping) is currently based on bit 1 of the flags. Might be wrong
5. Master status (t/f) is currently based on bit 2 of the flags. Might be wrong
6. Loadspec seems to determine what fastfile/zone the aliases are compiled into. So it is a pre-compilation flag which is not easy to reverse
7. Compression, reverb, and platform are likely present in the form of bit masked values of the 'flags' field. Need to research and test to confirm.

pagingred:
Should the contained SpeakerMap and SndCurve be dumped to their respective files? If so, do they need "registered" or something like that to distinguish them as assets?

pagingred:
> There is still significant work to be done on this feature. AFAIK the process for compiling these alias lists is lossy.
> 
>     1. The min and max volume are multiplied by the set volume modifier group's value. At this point I cannot discern what the original group was per sound alias (working on a heuristic) so I cannot determine the correct original volume.
> 
>     2. The volume modifier group itself, since it cannot be determined rn, is not dumped. Future games (like AW) have a separate field for this 'volModIndex' which makes me believe this value is present in the form of a bit masked value in the flags field.
> 
>     3. Alias type (loaded, streamed, primed) is currently based on bits 6-7 of the flags. Might be wrong
> 
>     4. Loop status (nonlooping, looping) is currently based on bit 1 of the flags. Might be wrong
> 
>     5. Master status (t/f) is currently based on bit 2 of the flags. Might be wrong
> 
>     6. Loadspec seems to determine what fastfile/zone the aliases are compiled into. So it is a pre-compilation flag which is not easy to reverse
> 
>     7. Compression, reverb, and platform are likely present in the form of bit masked values of the 'flags' field. Need to research and test to confirm.

Some updates to the previous points
1. Since I can not likely recover the original volmod and volumes, I just write the modified values and set the volmod to 'all'. This makes the multiplier 1.0 so there is no change from if it had the correct volume and unmodified volume.
2. As mentioned above, dumping 'all' by default for the volmod
3. Correctly changed to be based on bits 7-8. If bit 7, loaded. If but 8, streamed. If both, primed.
4. Loop status (nonlooping, looping) is still based on bit 1 of the flags. If bit 6 is set, loop status is rlooping.
5. Master value checks bit 2 and writes "master" if set. If not writes slavePercentage.
6. Loadspec left as empty since it only seemed to be used pre-compilation
7. Compression still unknown, not found in bits.
8. Reverb based on bits 4-5. If bit 4 set, add fulldrylevel if bit 5 set, add nowetlevel
9. Platform still unknown, not found in bits.

Raw binary research: https://pastebin.com/4AHeYPqx

pagingred:
> Should the contained SpeakerMap and SndCurve be dumped to their respective files? If so, do they need "registered" or something like that to distinguish them as assets?

For future reference:
The SpeakerMap is not a core asset type so it cannot be registered with the asset system. Therefore it requires explicit dumping and loading. The SndCurve however is a core asset type, so it does not require an explicit dumper/loader.

PR #768, opened 2026-04-30
feat: IW3 xanim dumping/loading in CoD4 Mod Tools raw binary format

michaeloliverx:
This PR adds IW3 `xanim` export to the CoD4 Mod Tools raw binary format.

It writes `xanim/<name>` directly from loaded `IW3::XAnimParts` data. The exporter reconstructs the raw header, delta data, quat/trans tracks, masks, bone names, and notify tail, and matches the linker’s quat packing and translation scaling behavior.

The goal is not to reproduce every stock raw file byte-for-byte. The CoD4 Mod Tools `linker_pc.exe` normalizes many `xanim` files into a canonical form. This exporter writes that canonical form. In practice that means stock raw files often change after one `linker_pc.exe -> unlink` pass, but files exported by this PR are stable after the first pass.

Verification was done against the real CoD4 Mod Tools pipeline using all `5258` stock `raw/xanim` [assets](https://github.com/promod/CoD4-Mod-Tools/tree/7e538c6632a5f2311a2a16f86e5b6786b4e65d16/raw/xanim):
- stock source vs first unlink dump: `339` exact, `4919` byte diffs, `0` size diffs
- roundtrip source vs second dump: `5258 / 5258` exact

So the output is 100% compatible with the CoD4 Mod Tools xanim binary format and is a stable fixed point under `linker_pc.exe -> unlink -> export`.

All the `viewmodel_` xanim have been tested in-game [JerryALT](https://gitea.com/JerryALT) by relinking the output here using the `linker_pc.exe` again.

**Edit**: Jerry has also verified _some_ delta anims and notetracks.


src/ObjWriting/Game/IW3/XAnim/XAnimDumperIW3.cpp
```diff
@@ -0,0 +1,912 @@
...
+        for (auto i = 0uz; i < nameCount; i++)
+            bones[i].name = ResolveScriptString(asset, parts.names[i]);
+
+        auto cursor = FlatDataCursor{
+            .dataByte = parts.dataByte,
+            .dataShort = parts.dataShort,
+            .dataInt = parts.dataInt,
+            .randomDataByte = parts.randomDataByte,
+            .randomDataShort = parts.randomDataShort,
+            .indices = parts.indices._2,
```
Laupetin, line 355: This is not correct though, right? The ZoneCode specifies `set condition indices::_1 numframes < 256;`, so when `useByteIndices` is set, this should use `_1` and read u8 instead of u16?
Or am I overlooking something that guarantees this is always the u16 variant here?

I did go through every vanilla IW3 fastfile and there is no anim that `useByteIndices && parts->indices.data` applies to. But for other games or custom anims this might, right? 🤔 

Edit: Okay no game actually has `useByteIndices` and a value in `parts->indices` and reading the code i can see why i guess? When `useByteIndices` it reads the indices from `byteData` instead of `indices`? Not sure what the whole thing about having this extra struct is then, but okay 😅

src/ObjWriting/Game/IW3/XAnim/XAnimDumperIW3.cpp
```diff
@@ -0,0 +1,912 @@
...
+
+        // The linker moves 16-bit frame indices into the top-level indices pool only when
+        // the in-memory stored size is at least 64, i.e. frameCount >= 65.
+        if (storedSize >= 64u)
+        {
+            for (auto i = 0uz; i < count; i++)
+                result[i] = cursor.indices[i];
+
+            cursor.indices += count;
+            cursor.dataShort += ((count - 2uz) >> 8u) + 2uz;
```
Laupetin: Any idea what the data here means? It's not relevant for the dumped xanim file?

michaeloliverx: For 16-bit frame indices with `storedSize >= 64`, the linker moves the full index list into `XAnimParts::indices`. It also stores a sparse set of "checkpoint" values for that track in `XAnimParts::dataShort`.

Those checkpoint values are copied from positions in the full index list: the first entry, then every 256th entry, and always the final entry. The final entry is included even when it does not land exactly on a 256-entry boundary. e.g.

```text
indices[0]
indices[256]
indices[512]
indices[768]
...
indices[last]
```

So the number of shorts to skip is:
```
((count - 2) / 256) + 2
```

where count is `storedSize + 1`, i.e. the number of frame index entries. We should update the code to use `/ 256` instead of `>> 8u` here to make this clearer.

These checkpoint values are not raw xanim payload. They are part of the loaded `XAnimParts` representation, likely to make runtime frame lookup cheaper by narrowing the search to a 256-entry block?

For dumping, we read the actual frame indices from `cursor.indices`, then advance `cursor.dataShort` over the checkpoint values so the next track starts at the correct position. The raw writer later emits the reconstructed index vector via `WriteIndicesIfNeeded`.

src/ObjLoading/Game/IW3/XAnim/XAnimLoaderIW3.cpp
```diff
@@ -0,0 +1,771 @@
...
+            FlatDataWriteCursor writeCursor;
+
+            std::vector<size_t> boneOrder(boneCount);
+            std::ranges::iota(boneOrder, 0);
+
+            std::ranges::sort(boneOrder,
+                              [&boneTracks](const size_t i0, const size_t i1)
+                              {
+                                  return std::to_underlying(boneTracks[i0].quat.type) < std::to_underlying(boneTracks[i1].quat.type);
+                              });
```
michaeloliverx, line 728: The linker comparator has a tie breaker, when two quat types are equal it compares the stored order byte e.g.
```c++
if (type0 == type1)
    return order0 - order1;
```

The same thing happens with trans.

I guess something like this?
```c++
// quat order
return std::tuple(quatType0, i0) < std::tuple(quatType1, i1);

// trans order
return std::tuple(transType0, partsBoneIndex0) < std::tuple(transType1, partsBoneIndex1);
```

src/ObjWriting/Game/IW3/XAnim/XAnimDumperIW3.cpp
```diff
@@ -0,0 +1,936 @@
...
+
+        const auto numLoopFrames = GetNumLoopFrames(*parts);
+        const auto useByteIndices = UseByteIndices(*parts);
+        auto maybeBoneTracks = ReconstructBoneTracks(asset);
+        const auto deltaTrack = ReconstructDeltaTrack(*parts);
+
+        if (!maybeBoneTracks.has_value())
+        {
+            con::error(maybeBoneTracks.error());
+            return;
```
michaeloliverx, line 872: Will this leave an empty file since the file has been opened already?

src/ObjLoading/Game/IW3/XAnim/XAnimLoaderIW3.cpp
```diff
@@ -0,0 +1,771 @@
...
+            for (auto notifyIndex = 0u; notifyIndex < numDiskNoteTracks; notifyIndex++)
+            {
+                auto& notify = parts.notify[notifyIndex];
+
+                const auto notifyName = stream::ReadCString(stream);
+                notify.name = m_script_strings.AddOrGetScriptString(notifyName);
+                registration.AddScriptString(notify.name);
+
+                const auto frame = stream::ReadValue<uint16_t>(stream);
+                notify.time = static_cast<float>(frame) / static_cast<float>(parts.numframes);
```
michaeloliverx: The linker does this:
```c++
frame = *rawFrame;
numframes = parts->numframes;

if (numframes)
    notify->time = (double)frame / (double)numframes;
else
    notify->time = 0.0;
```

Here OAT will divide by 0 if `parts.numframes` is `0`. Not sure if this is a real scenario though but it is a behaviour mismatch.

src/ObjLoading/Game/IW3/XAnim/XAnimLoaderIW3.cpp
Laupetin: that seems to be very important actually, without this the xanims from modtools linked with OAT do not match the xanims linked with official linker :o

src/ObjWriting/Game/IW3/XAnim/XAnimDumperIW3.cpp
Laupetin: Currently yes. In theory the error should only display on invalid xanims in the fastfile but yeah i guess i can move the bone track reconstruction before opening the file 🤔

PR #769, opened 2026-04-30
refactor(zcg): use wordsize instead of architecture

michaeloliverx:
`architecture x86/x64` in ZCG was really describing pointer width, not CPU architecture.

This renames that concept to `wordsize`, updates existing command files, and renames the related internal enum/APIs.

No behaviour is intended to change.


Laupetin (approved):
Very good change, makes a lot of sense considering xenon :D Thanks

PR #774, opened 2026-04-30
Add T5 GfxLightDef Dumping & Loading Logic

pagingred:
This PR adds in functionality to load and dump an T5 GfxLightDef. It was created as a 1:1 copy of the IW4 loader/dumper.


pagingred:
Moving the GfxLightDef PRs to a single one

PR #775, opened 2026-04-30
Add T6 GfxLightDef Dumping & Loading Logic

pagingred:
This PR adds in functionality to load and dump an T6 GfxLightDef. It was created as a 1:1 copy of the IW4 loader/dumper.


pagingred:
Failed the build so marking as draft

pagingred:
Moving the GfxLightDef PRs to a single one

Issue #327, opened 2025-01-03
Remove deprecated >type metadata in zone definitions


Laupetin:
Previously it has been possible to specify a type for a target in a zone definition via:
```
>type,fastfile
>type,ipak
>type,iwd
```

In the process of #286 this has been replaced by specifying obj containers like this:
```
>iwd,myiwd
>ipak,myipak
```

This allows obj containers to be created alongside a fastfile in one target.
Due to this, it is no longer necessary to have a `type` metadata.

For backwards compatibility the type property currently does the same thing as specifying the respective obj container with the name of the target.
It does log a deprecation warning however and will be removed in the future!

PR #777, opened 2026-05-01
fix: remove deprecated '>type,type' zone definition from examples

michaeloliverx:
Found it when playing with the `ExtendZone` example.

```
WARN: Deprecated: mp_showdown.zone L4: Zone definition ">type,type" is deprecated and should be removed. A fastfile will always be built when appropriate assets have been added.
```

Relates to https://github.com/Laupetin/OpenAssetTools/issues/327


Laupetin (approved):
Thanks :)

PR #779, opened 2026-05-01
feat: IW3-5, T5-6 fx dumping & loading logic

pagingred:
WORK IN PROGRESS

This PR implements logic to dump and load FxImpactTable and FxEffectDef assets from IW3, IW4, IW5, T5, and T6. 

IW3 FxImpactTable dumping is done, more to come soon.


src/ObjCommon/Game/IW3/FxImpactTable/FxImpactTableFields.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include "Game/IW3/IW3.h"
+
+#include <map>
+#include <utility>
+
+namespace IW3
+{
+    static constexpr auto FX_IMPACT_TABLE_ENTRY_COUNT = 12;
```
pagingred: Would it be better to use std::extent_v instead of hard-coding a constant?

Laupetin: yup, if possible that is always better as it makes it a single point of truth. When changing the entries it requires editing less places, it makes sure the value is never wrong and also documents what the number means.

pagingred:
Pivoted to also dump FxEffectDef. Some unknowns that are still in the works for IW3:
- Source of each FxElemDef's name is unknown. String instances were not found in ScriptStrings either. Not in struct AFAIK
- 'editorFlags' in source assets cannot be replicated. Not in struct AFAIK
- 'flags' in compiled assets is a 32 bit-packed integer. Here is current work:
```
        unsigned int unknown1 : 1;
        unsigned int spawnRelative : 1;
        unsigned int spawnFrustumCull : 1;
        unsigned int unknown2 : 4;
        unsigned int spawnOffsetNone : 2;
        unsigned int unknown3 : 16;
        unsigned int runRelToEffect : 1;
        unsigned int runRelToSpawn : 1;
        unsigned int unknown4 : 1;
        unsigned int nonUniformScale : 1;
        unsigned int unknown5 : 3;
```
- 'atlas.behavior' in compiled assets is an 8-bit packed integer. Here is current work:
```
        unsigned int startRandom : 1;
        unsigned int startIndexed : 1;
        unsigned int playOverLife : 1;
        unsigned int loopOnlyNTimes : 1;
        unsigned int unknown: 4;
```
- Both 'spawnLooping' and 'spawnOneShot' are in source asset, yet internal union can only support one at a time. Possibly based on a bit flag.
- No clue where velGraph[0,1][X,Y,Z], rotGraph, sizeGraph[0,1], scaleGraph, colorGraph, alphaGraph come from. Likely visSamples and velSamples.
- 'collOffset' likely is derived from the collMin and collMax. Set to collMin for now
- 'collRadius' origin unknown, likely is derived from the collMin and collMax.

PR #781, opened 2026-05-01
fix: omitting default armature causes problems

Laupetin:
There are two problems that result from the logic that determines whether a default armature can be omitted:
1. The previous assumption that all verts and tris are either part of a vertlist or vertblend was wrong, IW3 seems to have some models that have one tri missing from the only vertlist. It kind of feels like a bug? But it makes the current logic not work. Also since it only occurs on LOD 0, the lower lods have their armature omitted leading to bones not matching between them when loading
2. The name of the root bone follows not particular pattern and varies from model to model. It seems like it also doesn't match the xmodel name often times. In some instances, the name of the root bone also matters, e.g. when attaching one xmodel to another, see #715. So the name of the root bone cannot change, even when omitting the default armature.

fixes: #715

### DIFF
--- a/src/Common/Game/IW3/IW3_Assets.h
+++ b/src/Common/Game/IW3/IW3_Assets.h
@@ -530,7 +530,7 @@ namespace IW3
         int16_t collLod;
         XModelStreamInfo streamInfo;
         int memUsage;
-        char flags;
+        unsigned char flags;
         bool bad;
         PhysPreset* physPreset;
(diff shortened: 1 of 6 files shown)


Laupetin:
Instead of getting rid of omitted armatures i decided to add an optional property in the xmodel json instead: `rootBoneName`.
This will make it possible to override the name of the root bone of the generated armature.
As the default armature only has one bone, it is possible to for only one name to be specified.

PR #797, opened 2026-05-11
feat: ipak dumping without ff

Laupetin:
As this is currently not possible but there has been interest in doing something like this from time to time, this PR makes ImageConverter able to also just dump all data from an ipak.

The images will not be named, as the names can only be known when the image data is referenced from a fastfile.
But this is still useful when dealing with console ipaks and scanning ipaks for data that is unused.

TODO:
- [x] xenon ipaks support XMemDecompress which PC doesn't
- [x] only set .iwi extension when its an iwi


Laupetin:
Okay so xenon ipak dumping works but what I didn't know yet is that they do _not_ contain iwis 😅
The data inside is probably raw texture data that is even separated by mips or something along this line.

I'm afraid that makes dumping xenon ipaks not very useful for now as the fastfiles contain the necessary metadata of texture format, dimension and potentially more that is required for making sense of the data.

PR #800, opened 2026-05-14
Fix IW5 x64 Zone Loading

pagingred:
Seems like most assets call to `MaybePointerFromLookup(void* ptr)` when resolving an internal pointer. This constructor variant sets m_valid to true by default. Since the loader calls `Expect()` this flag MUST be true to avoid throwing an exception.

When loading the brushes for a clipmap, it uses the `MaybePointerFromLookup(void* ptr, const block_t block, const size_t offset)` variant instead. For some reason this one had set m_valid to false instead of true, so this PR just flips that.

### DIFF
--- a/src/ZoneLoading/Zone/Stream/ZoneInputStream.h
+++ b/src/ZoneLoading/Zone/Stream/ZoneInputStream.h
@@ -68,7 +68,7 @@ public:
     }
 
     MaybePointerFromLookup(void* ptr, const block_t block, const size_t offset)
-        : m_valid(false),
+        : m_valid(true),
           m_ptr(ptr),
           m_block(block),
           m_offset(offset)


Laupetin (commented):
hey, thanks for putting in time to investigate this. Yeah it is true that this makes it so it does not crash anymore, but it does not solve the underlying issue either unfortunately 😬 

The issue in `berlin.ff` is encountered when the `ClipInfo` struct is loaded a second time and the `brushes` member is loaded.
The first time, the data is follow, the second time it is a reference.
If you compare the data that is being loaded between the native x86 version and the x64 version when it references the offset the second time, you will notice that the data does not match and the data in the x64 version is invalid (invalid pointer with `sides`).

In general this cannot reference block memory as the `cbrush_t` struct contains a pointer which makes the struct incompatible between x86 and x64.

When loading you will not notice that the pointer to the `cbrush_t` array is invalid because it is never read again.
Any future dumper of this struct or trying to write the clipmap again from in-memory will result in a crash however.

So the issue is deeper. I agree though that this place in code is a bit confusing, i don't know either why valid is set to false there but still the block memory members are set. I should have maybe commented something there so i could tell you why, but i don't remember.

## What x64 loads:



## What x86 loads and what the data should actually look like:

PR #801, opened 2026-05-15
fix(unlinker): respect `--minimal-zone` CLI arg

michaeloliverx:


Laupetin (approved):
Not so much a fix but more actually implementing the feature 😅 
I did a small commit that extracts the DependencyAssetLookup to a separate class and renames it to "ZoneDefFilter" to make it more agnostic to how it is filtering.
Apprechiate the PR, thank you very much :)

PR #807, opened 2026-05-19
feat: initial T4 support

michaeloliverx:
Initial T4 support which focuses only on the unlinking / parsing side for now. I added a few of the simple dumpers rawfile,images and localize. Obj loading can be done at a later stage for easier review here. This PR can successfully unlink all retail T4 PC fastfiles.

I initially duplicated IW3 and then modified it so there may be some leftovers I have forgotten.

I have NOT tested x64 yet but this should be enough to get initial review comments :)

The T4 asset structs come from https://github.com/JezuzLizard/T4SP-Server-Plugin/blob/main/src/game/xasset.hpp, I got permission from one of the authors @ineedbots to use that file.


src/ObjWriting/Game/T4/Localize/LocalizeDumperT4.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+        const auto language = LocalizeCommon::GetNameOfLanguage(context.m_zone.m_language);
+        const auto assetFile = context.OpenAssetFile(std::format("{}/localizedstrings/{}.str", language, context.m_zone.m_name));
+
+        if (assetFile)
+        {
+            StringFileDumper stringFileDumper(context.m_zone, *assetFile);
+
+            stringFileDumper.SetLanguageName(language);
+            stringFileDumper.SetConfigFile(R"(G:\CoD5\cod\cod5\bin\StringEd.cfg)");
```
michaeloliverx, line 27: Came from the cod5 mod tools. https://code.idtech.space/treyarch/cod5-sdk/src/branch/master/raw/english/localizedstrings/anim_test.str#L13

src/ZoneCode/Game/T4/XAssets/XModel.txt
```diff
@@ -0,0 +1,90 @@
...
+set reusable partClassification;
+set count partClassification numBones;
+set reusable baseMat;
+set count baseMat numBones;
+set count surfs numsurfs;
+set count materialHandles numsurfs;
+set count collSurfs numCollSurfs;
+set count boneInfo numBones;
+// The DB loader does not follow XModelStreamInfo::highMipBounds here.
+set condition streamInfo::highMipBounds never;
```
michaeloliverx: This is strange right?

pagingred: T5 seems to allocate an array with size = numsurfs

`set count streamInfo::highMipBounds numsurfs;`

Laupetin: I checked with every asset on the vanilla fastfiles and this struct is always entirely nulled.
Cod4 has this:
```cpp
struct XModelStreamInfo
{
	char pad;
};
```
so i would assume the struct here is just the same as cod4?
T4 should be based on the IW3 engine so chances are a lot of things are similar?

Together with the alignment C does here it's also 4 bytes.
Even if it wasn't like cod4, doesn't seem to make a difference, whether you have a pointer that is never loaded or a struct with only padding.

I'd probably prefer just adjusting it to cod4 style as this eliminates the chance somebody in the future tries to add actual data here or think there has to be actual data.

The game does not have this in its PC loading code.

src/Common/Game/T4/CommonT4.cpp
```diff
@@ -0,0 +1,35 @@
+#include "CommonT4.h"
```
michaeloliverx, line 1: Most if not all of these are unused atm? from copying IW3.

src/ZoneCode/Game/T4/XAssets/XModel.txt
michaeloliverx: > T4 should be based on the IW3 engine so chances are a lot of things are similar?

Yes, definitely.

This struct exists in the Xenon builds of both IW3 and T4, and the engine does actually load it there. I suspect it’s just a quirk of their asset/build pipeline or how the game was built internally.

src/Common/Game/T4/CommonT4.cpp
Laupetin: that's cool i guess. Goal is feature parity anyway so those will be needed at some point and knowing what algorithms T4 uses has a value.

src/Common/Game/T4/GameT4.cpp
```diff
@@ -0,0 +1,96 @@
...
+        "destructibledef",
+        "xanim",
+        "xmodel",
+        "material",
+        "techset",
+        "image",
+        "sound",
+        "loaded_sound",
+        "col_map_sp",
+        "col_map_mp",
```
Laupetin: yeah i know technically this is what the game is officially calling it 😅 but that differs between the games and i like having the same name across games.
So i'd call this:
```
"clipmap_unused",
"clipmap",
```
(given that the "sp" asset is actually unused, like in all other games)

Also other assets i'd adapt the commonly used name across other games:
```
techniqueset
loadedsound
comworld
gameworldsp
gameworldmp
mapents
gfxworld
uimap
menulist
```

If supporting official mod tools names is important to you: They can be aliases.
I noticed some of the ones that are currently there are duplicates of the names you currently have already.
The main names should be the ones that stay the same across games though as those are picked when dumping zone files and logging.

src/Common/Game/T4/T4.h
```diff
@@ -0,0 +1,150 @@
...
+    };
+
+    struct XAssetList
+    {
+        ScriptStringList stringList;
+        int assetCount;
+        XAsset* assets;
+    };
+
+    using AssetXModelPieces = Asset<ASSET_TYPE_XMODELPIECES, XModelPieces>;
```
Laupetin: on the other games i only include the games as assets that are independently loaded from XAssetHeader (and are therefore marked as assets in T4.txt).

<sub><i>i seemingly just forgot to remove it in IW3, whoops</i></sub>

src/Common/Game/T4/T4.h
```diff
@@ -0,0 +1,150 @@
...
+    using AssetMenuList = Asset<ASSET_TYPE_MENULIST, MenuList>;
+    using AssetMenu = Asset<ASSET_TYPE_MENU, menuDef_t>;
+    using AssetLocalize = Asset<ASSET_TYPE_LOCALIZE_ENTRY, LocalizeEntry>;
+    using AssetWeapon = Asset<ASSET_TYPE_WEAPON, WeaponDef>;
+    using AssetSoundDriverGlobals = Asset<ASSET_TYPE_SNDDRIVER_GLOBALS, SndDriverGlobals>;
+    using AssetFx = Asset<ASSET_TYPE_FX, FxEffectDef>;
+    using AssetImpactFx = Asset<ASSET_TYPE_IMPACT_FX, FxImpactTable>;
+    using AssetRawFile = Asset<ASSET_TYPE_RAWFILE, RawFile>;
+    using AssetStringTable = Asset<ASSET_TYPE_STRINGTABLE, StringTable>;
+
```
Laupetin, line 115: This is missing `AssetPackIndex`. That one is actually required as the game can load this asset type on PC.

src/ObjCompiling/IObjCompiler.cpp
```diff
@@ -12,6 +12,7 @@ const IObjCompiler* IObjCompiler::GetObjCompilerForGame(GameId game)
 {
     static const IObjCompiler* objCompilers[static_cast<unsigned>(GameId::COUNT)]{
         new IW3::ObjCompiler(),
+        nullptr,
```
Laupetin: I'd also implement a dummy instance here similar to ObjLoader, because it's easy to forgot here

src/ObjLoading/Game/T4/ObjLoaderT4.cpp
```diff
@@ -0,0 +1,9 @@
+#include "ObjLoaderT4.h"
+
+using namespace T4;
+
+void ObjLoader::LoadReferencedContainersForZone(ISearchPath& searchPath, Zone& zone) const {}
+
+void ObjLoader::UnloadContainersOfZone(Zone& zone) const {}
+
+void ObjLoader::ConfigureCreatorCollection(AssetCreatorCollection& collection, Zone& zone, ISearchPath& searchPath, IGdtQueryable& gdt) const {}
```
Laupetin: I'd still add `GlobalAssetPoolsLoader` and `DefaultAssetCreator` instances here i guess as that's boilerplate anyway

src/ObjWriting/Image/ImageDumper.cpp.template
```diff
@@ -1,10 +1,10 @@
-#options GAME (IW3, IW4, IW5, T5, T6)
+#options GAME (IW3, T4, IW4, IW5, T5, T6)
 
 #filename "Game/" + GAME + "/Image/ImageDumper" + GAME + ".cpp"
 
 #set DUMPER_HEADER "\"ImageDumper" + GAME + ".h\""
 
-#if GAME == "IW3"
+#if GAME == "IW3" || GAME == "T4"
```
Laupetin: The game should have its own section as it defines `FEATURE_IW3` for T4 otherwise 😅

src/ZoneLoading/Game/T4/XAssets/loadedsound/loadedsound_actions.cpp
Laupetin, line 1: Other games have LoadedSound data in the temp block, so it has to be reallocated.
T4 does not seem to have its sound data in the temp block, so i guess this action can be omitted entirely.

src/ZoneLoading/Game/T4/ContentLoaderT4.cpp
```diff
@@ -0,0 +1,178 @@
...
+    assert(varXAsset != nullptr);
+
+    if (atStreamStart)
+        m_stream.Load<XAsset>(varXAsset);
+
+    const auto assetTypeName = IGame::GetGameById(m_zone.m_game_id)->GetAssetTypeName(varXAsset->type);
+    if (assetTypeName.has_value())
+        con::debug(R"(Loading {})", *assetTypeName);
+    else
+        con::debug("Loading asset type {}", static_cast<int>(varXAsset->type));
```
Laupetin: There seems to be some leftover debugging code 😛

src/ZoneLoading/Game/T4/ContentLoaderT4.cpp
```diff
@@ -0,0 +1,178 @@
...
+        LOAD_ASSET(ASSET_TYPE_SNDDRIVER_GLOBALS, SndDriverGlobals, sndDriverGlobals)
+        LOAD_ASSET(ASSET_TYPE_FX, FxEffectDef, fx)
+        LOAD_ASSET(ASSET_TYPE_IMPACT_FX, FxImpactTable, impactFx)
+        SKIP_ASSET(ASSET_TYPE_AITYPE, AiType, data)
+        SKIP_ASSET(ASSET_TYPE_MPTYPE, MpType, data)
+        SKIP_ASSET(ASSET_TYPE_CHARACTER, Character, data)
+        SKIP_ASSET(ASSET_TYPE_XMODELALIAS, XModelAlias, data)
+        LOAD_ASSET(ASSET_TYPE_RAWFILE, RawFile, rawfile)
+        LOAD_ASSET(ASSET_TYPE_STRINGTABLE, StringTable, stringTable)
+        SKIP_ASSET(ASSET_TYPE_PACK_INDEX, PackIndex, data)
```
Laupetin: That is not accurate: The game actually loads PackIndex assets.

src/ZoneCode/Game/T4/XAssets/LoadedSound.txt
```diff
@@ -0,0 +1,12 @@
+// =========================================
+// LoadedSound
+// =========================================
+use LoadedSound;
+set block XFILE_BLOCK_TEMP;
+set string name;
+
+// snd_asset
+use snd_asset;
+set reusable data;
```
Laupetin: `data` does not seem to be reusable?

src/ZoneCode/Game/T4/XAssets/menuDef_t.txt
```diff
@@ -0,0 +1,81 @@
...
+set string mouseExit;
+set string action;
+set string onAccept;
+set string onFocus;
+set string leaveFocus;
+set string dvar;
+set string dvarTest;
+set string onListboxSelectionChange;
+set string enableDvar;
+set condition focusSound never;
```
Laupetin: Why? Game's ZoneCode says this is a `snd_alias_list_t*`, just like IW3?

src/ZoneWriting/Writing/IZoneWriterFactory.cpp
```diff
@@ -12,6 +12,7 @@ const IZoneWriterFactory* IZoneWriterFactory::GetZoneWriterFactoryForGame(GameId
 {
     static const IZoneWriterFactory* zoneCreators[static_cast<unsigned>(GameId::COUNT)]{
         new IW3::ZoneWriterFactory(),
+        nullptr,
```
Laupetin: I apprechiate the effort to keep the PR small for better reviewability :D I think I will add a zone writer factory here though to prevent unexpected crashes 😅

Laupetin (commented):
Thanks for putting in the effort to make the initial version work :D I'm adding a few fixes on top.
I can unlink every WaW fastfile with this initial version successfully (though I only have the german version on Steam) in both x86 and x64.

src/Common/Game/T4/T4_Assets.h
```diff
@@ -0,0 +1,4109 @@
...
+        STATE_FLAG_CULL_FRONT = 0x2,
+        STATE_FLAG_DECAL = 0x4,
+        STATE_FLAG_WRITES_DEPTH = 0x8,
+        STATE_FLAG_USES_DEPTH_BUFFER = 0x10,
+        STATE_FLAG_USES_STENCIL_BUFFER = 0x20,
+        STATE_FLAG_CULL_BACK_SHADOW = 0x40,
+        STATE_FLAG_CULL_FRONT_SHADOW = 0x80,
+    };
+
+    struct Material
```
Laupetin, line 1072: After #823 is merged this should be

```suggestion
    struct zcg_align(4) Material
```

src/Common/Game/IGame.h
Laupetin, line 22: I know what you did there, you wanted to order according to release date.
But it's actually sorted alphabetically 😅 As otherwise T5 would have to be between IW4 and IW5

src/Common/Game/T4/T4_Assets.h
Laupetin: nvm, that didn't work, since the struct tests fail as the expected alignment differs from the one from the compiler.
So it's `set allocalign Material 4;` instead in `Material.txt`

src/ZoneCode/Game/T4/XAssets/LoadedSound.txt
michaeloliverx: Yep looks like a mistake.

src/Common/Game/T4/GameT4.cpp
michaeloliverx: This is fine by me, I just assumed they matched the assetNames array exactly in the executable binary.

Laupetin:
Rebuilt all vanilla fastfiles and dumped them again with both x86 and x64 and it seems to work fine. With the last fix required merged with #825 i think this should be good to go. Thank you very much @michaeloliverx :)

PR #808, opened 2026-05-21
feat: Initial IW6

pagingred:
This PR is still in progress. Most of the necessary logic was ported from IW5, and cross-referenced to IDA. Should support the same dumpers as IW5 with minimal additional work.

I believe I have most of the assets worked out and from the correct sources, however I am working through some pointer issues regarding 64bit v.s. 32bit.


Laupetin:
I haven't look at this yet (and it will probably take me a bit as well as there's a lot on the todo list 😅) but as a heads up: This PR is *very* large, maybe consider breaking this up. 

And also: I would not implement any x86 support for iw6, that would require _a lot_ of abstraction in the fastfile code generation similar to how x86 on x64 hosts work but it's not giving any real benefit. 

I'm planning on switching the main release version to x64 at some point when the x64 version is battle proven and tested well for x86 games and works good.
The x86 version will then only be kept around for cross-checking issues with a native loading process.

So i'd just not support ghosts on x86 hosts.

pagingred:
> I haven't look at this yet (and it will probably take me a bit as well as there's a lot on the todo list 😅) but as a heads up: This PR is *very* large, maybe consider breaking this up. 
> 
> 
> 
> And also: I would not implement any x86 support for iw6, that would require _a lot_ of abstraction in the fastfile code generation similar to how x86 on x64 hosts work but it's not giving any real benefit. 
> 
> 
> 
> I'm planning on switching the main release version to x64 at some point when the x64 version is battle proven and tested well for x86 games and works good.
> 
> The x86 version will then only be kept around for cross-checking issues with a native loading process.
> 
> 
> 
> So i'd just not support ghosts on x86 hosts.

Understandable. I was going for a 1:1 port of the IW5 work but I can see that was a bit ambitious. I'll see how I can break this up into multiple sets of changes. Thanks

pagingred:
Closing until this can be broken down into smaller sets of changes.

PR #818, opened 2026-06-05
feat: add binary xanim support for remaining games

Laupetin:
Add support for xanim in the official modding tools binary format:
* IW3: Version 17
* T4: Version 17
* IW4: Version ?
* IW5: Version ?
* T5: Version 19
* T6: Version 19 (but the format is not identical to T5...)

Might make up a similar format for IW4 and IW5 and make it version number 18.
There's no way to know whether that's exactly correct but i mean if there is no official mod tools, it can't really be incompatible with something official anyway.

Note: Not that this is relevant for this project but bo3 seems to use version 40 so it is not like the version 18 comes up in later mod tools.


Laupetin:
I ended up also using version number 19 for T6. However there is not an additional flag `0x80` that specifies that the flags and the rest should be parsed as T6 would do, rather than like T5.

Also T6 dumps all anims that do not make use of any T6 exclusive feature (so delta3D) in T5's format to make it compatible with T5 official modding tools.
Same goes for IW4 and IW5 which dump their anims in IW3's version 17 when there is no delta3D involved, so their anims can be used in IW3's official modding tools as much as possible.

OAT itself can load anims from all versions.
The only thing that is not supported is loading an animation with delta3D in a game that does not support delta3D.

Cowldwink:
Absolute cinema 🙌

uhhashe:
I remember seeing references to streamed xanims, wonder what it was used for
Existed in both T5 and T6 code iirc

Laupetin:
> I remember seeing references to streamed xanims, wonder what it was used for Existed in both T5 and T6 code iirc

Having seen how images work on xenon recently my guess would be that this is a console exclusive optimization?
At least there does not seem to be a single T5 xanim that uses the `STREAMED` flag and on T6 it seems to no longer exist on PC.

For images they separated miplevels into separate entries that can be streamed in at runtime instead of always loading the entire image with all miplevels that are needed like PC.
So maybe on console they can also stream in animation data on demand from pak files instead of always having them in-memory.

PR #826, opened 2026-06-11
Add dumping support for PS3, Wii U and Xenon server fast files

uhhashe:
Added some basic checks to allow dumping PS3, Wii U and Xenon server fast files.
From my research they're just the same as PC server fast files but with a different version number

### DIFF
--- a/src/Common/Game/IGame.h
+++ b/src/Common/Game/IGame.h
@@ -42,7 +42,8 @@ enum class GamePlatform : std::uint8_t
 {
     PC,
     XBOX,
-    PS3
+    PS3,
+    WIIU
 };
 
 static constexpr const char* GameId_Names[]{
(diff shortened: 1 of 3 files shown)


Laupetin:
Ehm okay that would require that all of those games on these platform use exactly 100% identical structs on all assets? Including all identical alignments and everything? That's for sure not the case on Xbox? Did you to through the zone loading code of each of those games? Or did you check whether you could actually load and dump all of the fastfiles of these games?

uhhashe:
It's for server fast files, which all run on PC and are in the TAsvu100 format
I did a couple tests with the fast files and they all dumped fine
rawfile, xmodel, material, animtree, stringtable were some of the assets I dumped
Should be 1:1 anyways
Reason I made this PR was so people didn't have to manually edit the version number to dump platform server fast files

Laupetin (approved):
alright i see, thanks :)

PR #827, opened 2026-06-11
docs: adjust xanim support for iw4,iw5,t5,t6

uhhashe:
Added ticks to XAnimParts dumper to all games except T4


Laupetin (approved):
Guess I forgot that, thanks 👍

PR #842, opened 2026-06-20
feat: add addon map ents dumper to T6

uhhashe:
Adds the abillity to dump `AddonMapEnts` assets from T6.
Zombie gametype ents and spec ops (Strike Force) ents from Campaign are now dumpable

### DIFF
--- a/src/ObjWriting/Game/T6/Maps/AddonMapEntsDumperT6.cpp
+++ b/src/ObjWriting/Game/T6/Maps/AddonMapEntsDumperT6.cpp
@@ -1,4 +1,3 @@
-#define NOMINMAX
 #include "AddonMapEntsDumperT6.h"
 
 #include <algorithm>
--- a/src/ObjWriting/Game/T6/Maps/AddonMapEntsDumperT6.cpp
+++ b/src/ObjWriting/Game/T6/Maps/AddonMapEntsDumperT6.cpp
@@ -1,0 +1,21 @@
+#define NOMINMAX
+#include "AddonMapEntsDumperT6.h"
+
+#include <algorithm>
+
+using namespace T6;
+
+namespace addon_map_ents
+{
+    void DumperT6::DumpAsset(AssetDumpingContext& context, const XAssetInfo<AssetAddonMapEnts::Type>& asset)
+    {
+        const auto* addonMapEnts = asset.Asset();
+        const auto assetFile = context.OpenAssetFile(asset.m_name);
+
+        if (!assetFile)
+            return;
+
+        auto& stream = *assetFile;
+        stream.write(addonMapEnts->entityString, std::max(addonMapEnts->numEntityChars - 1, 0));
+    }
+} // namespace addon_map_ents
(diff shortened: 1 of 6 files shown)


src/ObjWriting/Game/T6/Maps/AddonMapEntsDumperT6.cpp
```diff
@@ -0,0 +1,21 @@
+#define NOMINMAX
```
Laupetin: That should not be necessary. The min and max macros are usually defined by the `windows.h` header, which should not be included anywhere here.

src/ObjWriting/Game/T6/Maps/AddonMapEntsDumperT6.cpp
```diff
@@ -0,0 +1,21 @@
...
+    void DumperT6::DumpAsset(AssetDumpingContext& context, const XAssetInfo<AssetAddonMapEnts::Type>& asset)
+    {
+        const auto* addonMapEnts = asset.Asset();
+        const auto assetFile = context.OpenAssetFile(asset.m_name);
+
+        if (!assetFile)
+            return;
+
+        auto& stream = *assetFile;
+        stream.write(addonMapEnts->entityString, std::max(addonMapEnts->numEntityChars - 1, 0));
```
Laupetin, line 18: The big issue is that this asset is more data than just the entity string.

e.g.
Image

Just dumping the entity string is not necessarly very helpful as you cannot reconstruct the asset from just that.
So it's not possible to create a new addonmapents asset from disk like that.

src/ObjWriting/Game/T6/Maps/AddonMapEntsDumperT6.cpp
uhhashe: Hey so I based this off the IW5 one, so it should also be removed from the others too

src/ObjWriting/Game/T6/Maps/AddonMapEntsDumperT6.cpp
uhhashe: Since Plutonium could load them from disk (assuming it's only entity string), I thought it would be useful for modders to be able to dump it alongside with map ents. I am aware Plutonium can dump them with `dump_assets` dvar

src/ObjWriting/Game/T6/Maps/AddonMapEntsDumperT6.cpp
Laupetin: yeah you are right, thanks 👍

src/ObjWriting/Game/T6/Maps/AddonMapEntsDumperT6.cpp
Laupetin: hm yeah, for some reason i thought that the games that _do_ have an addonmapents dumper currently just have addonmapents that _only_ have the string and no additional data.
But that's not the case. Neither for IW4 nor for IW5.

So yeah I guess that's a fair point.
Will push the fix for the other dumpers on the branch as well then, thanks :)

edit: ah you already did, nvm then

PR #852, opened 2026-06-25
feat: T4 XModel dumper

michaeloliverx:
Checked a few items and it looks okay.

image

Also works in ModMan 
image


Laupetin (approved):
thank you :)

PR #854, opened 2026-06-25
feat: T4 xanim binary dumping and loading

michaeloliverx:
Easy win as the format is identical to IW3 (17).


Laupetin (approved):
thanks :)

PR #855, opened 2026-06-25
feat: T4 LoadedSound dumper

michaeloliverx:
There seems to be another format I don't know what it is so skipped it for now.

### DIFF
--- a/src/ObjWriting/Game/T4/Sound/LoadedSoundDumperT4.cpp
+++ b/src/ObjWriting/Game/T4/Sound/LoadedSoundDumperT4.cpp
@@ -1,352 +1,66 @@
 #include "LoadedSoundDumperT4.h"
 
-#include "Utils/Endianness.h"
 #include "Utils/Logging/Log.h"
-#include "Utils/StreamUtils.h"
 
-#include <algorithm>
-#include <array>
-#include <cstddef>
-#include <cstdint>
-#include <expected>
-#include <format>
-#include <limits>
-#include <ostream>
+#include <cassert>
+#include <filesystem>
+#include <optional>
 #include <string>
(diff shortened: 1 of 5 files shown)


src/ObjWriting/Game/T4/Sound/LoadedSoundDumperT4.cpp
```diff
@@ -0,0 +1,392 @@
...
+        if (wavInfo.format == WavFormat::MsAdpcm)
+        {
+            const auto decoded = DecodeMsAdpcm(wavInfo);
+            if (!decoded.has_value())
+            {
+                con::warn("Could not decode MS ADPCM loaded sound '{}': {}", loadedSound->name, decoded.error());
+                return;
+            }
+
+            DumpDecodedMsAdpcm(context, asset, decoded.value());
```
Laupetin: I mean, it's nice that you could parse all of the wav headers and decode everything and stuff but i'm confused to why?

You can just do
```cpp
            const auto assetFile = context.OpenAssetFile(std::format("sound/{}", asset.m_name));
            if (!assetFile)
                return;

            assetFile->write(data, static_cast<std::streamsize>(dataSize));
```
no?

Isn't this overengineered?

michaeloliverx: That is a lot simpler 😂 
I don't know why I thought I had to output uncompressed PCM WAV..

Laupetin: Okay well then i stripped all the wav parsing code 😅 Sorry 😬 but if needed this code can be referenced from this PR

PR #850, opened 2026-06-24
feat: T6 font_s dumper

uhhashe:
This PR allows dumping font files for T6. This is mostly meant as a continuation of PR https://github.com/Laupetin/OpenAssetTools/pull/739 as that lacks T6 support

I have fixed a few things in that PR
- Removed the name field as it's not needed
- Fixed T6 Glyph types
- Added proper letter decoding from my own [tool](https://github.com/meowlyny/fontbin2json/blob/bd6d090315820598d40d0790b1d030fd318f5a24/fontbin2json.cpp#L101)

Let me know if anything needs changing or adding

PR #866, opened 2026-06-29
feat: font loading and dumping

michaeloliverx:
Relates to https://github.com/Laupetin/OpenAssetTools/pull/739 but that PR appears to be hanging around 🤷🏻 

Intentionally left out T6 as there is some work being done on that in https://github.com/Laupetin/OpenAssetTools/pull/850.

### DIFF
--- a/src/ObjCommon/Font/JsonFont.h.template
+++ b/src/ObjCommon/Font/JsonFont.h.template
@@ -8,10 +8,8 @@
 
 #pragma once
 
-#set GAME_HEADER "\"Game/" + GAME + "/" + GAME + ".h\""
-#include GAME_HEADER
-
 #include "Json/JsonExtension.h"
+#include "Font/FontCommon.h"
 
 #include <nlohmann/json.hpp>
 #include <optional>
@@ -23,7 +21,7 @@ namespace GAME
     class JsonGlyph
     {
     public:
-        int letter;
+        unsigned letter;
         int x0;
         int y0;
         unsigned dx;
@@ -35,7 +33,26 @@ namespace GAME
         float t1;
     };
 
-    NLOHMANN_DEFINE_TYPE_EXTENSION(JsonGlyph, letter, x0, y0, dx, pixelWidth, pixelHeight, s0, t0, s1, t1);
+    NLOHMANN_TO_JSON_METHOD(JsonGlyph)
+    {
+        if (font::IsPrintableLetter(nlohmann_json_t.letter))
+            extended_to_json("letter", nlohmann_json_j, font::LetterToString(nlohmann_json_t.letter));
+        else
+            extended_to_json("letter", nlohmann_json_j, nlohmann_json_t.letter);
+        
+        NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(EXTEND_JSON_TO, x0, y0, dx, pixelWidth, pixelHeight, s0, t0, s1, t1))
+    }
+    
+    NLOHMANN_FROM_JSON_METHOD(JsonGlyph)
+    {
+        const auto jLetter = nlohmann_json_j.at("letter");
+        if (jLetter.type() == BasicJsonType::value_t::string)
+            nlohmann_json_t.letter = font::StringToLetter(jLetter.template get<std::string>());
+        else
+            nlohmann_json_t.letter = jLetter.template get<unsigned>();
+            
+        NLOHMANN_JSON_EXPAND(NLOHMANN_JSON_PASTE(EXTEND_JSON_FROM, x0, y0, dx, pixelWidth, pixelHeight, s0, t0, s1, t1))
+    }
 
     class JsonFont
     {
--- a/src/ObjCommon/Font/JsonFont.h.template
+++ b/src/ObjCommon/Font/JsonFont.h.template
@@ -1,0 +1,51 @@
+#options GAME (IW3, IW4, IW5, T4, T5)
+
+#filename "Game/" + GAME + "/Font/JsonFont" + GAME + ".h"
+
+// This file was templated.
+// See JsonFont.h.template.
+// Do not modify, changes will be lost.
+
+#pragma once
+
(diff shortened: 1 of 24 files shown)


src/ObjWriting/Font/FontDumper.cpp.template
```diff
@@ -0,0 +1,128 @@
...
+            JsonFont jFont;
+            CreateJsonFont(jFont, font);
+
+            ordered_json jRoot;
+            jRoot["$schema"] = "http://openassettools.dev/schema/font.v1.json";
+            jRoot["_type"] = "font";
+            jRoot["_version"] = 1;
+            jRoot["_game"] = GAME_LOWER;
+            jRoot["pixelHeight"] = jFont.pixelHeight;
+            jRoot["glyphCount"] = jFont.glyphCount;
```
michaeloliverx: Should we write this key? Since it can be inferred from the size of the `glyphs` array?

Laupetin: What I did previously in other assets and would will continue to do is to leave away this kind of redundant information. If you want to change a fact about an asset, it's not very user friendly to have to change different fields in the same way.
So if it's so redundant like here i'd just calculate the count from the array.

src/ObjCommon/Font/JsonFont.h.template
```diff
@@ -0,0 +1,51 @@
...
+    };
+
+    NLOHMANN_DEFINE_TYPE_EXTENSION(JsonGlyph, letter, x0, y0, dx, pixelWidth, pixelHeight, s0, t0, s1, t1);
+
+    class JsonFont
+    {
+    public:
+        int pixelHeight;
+        int glyphCount;
+        std::optional<std::string> material;
```
Laupetin: i don't think the material is optional tbh

src/ObjCommon/Font/JsonFont.h.template
```diff
@@ -0,0 +1,51 @@
...
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace GAME
+{
+    class JsonGlyph
+    {
+    public:
+        int letter;
```
Laupetin: i like what @ashingware did in #850 -> making this a utf-8 string, so i'd do this here as well i guess

src/ObjWriting/Font/FontDumper.cpp.template
```diff
@@ -0,0 +1,128 @@
...
+
+            if (jFont.material)
+                jRoot["material"] = *jFont.material;
+
+            if (jFont.glowMaterial)
+                jRoot["glowMaterial"] = *jFont.glowMaterial;
+
+            jRoot["glyphs"] = jFont.glyphs;
```
Laupetin:
```suggestion
           jRoot.merge_patch(jFont);
```

No need to manually add the entries from the json struct 😛

src/ObjCommon/Font/JsonFont.h.template
```diff
@@ -0,0 +1,51 @@
...
+        int x0;
+        int y0;
+        int dx;
+        int pixelWidth;
+        int pixelHeight;
```
Laupetin: These 3 (`dx`, `pixelWidth`, `pixelHeight`) seem to be `unsigned char` in the games, so they can be `unsigned` here in the json struct as well

src/ObjCommon/Font/JsonFont.h.template
```diff
@@ -0,0 +1,51 @@
...
+        float s1;
+        float t1;
+    };
+
+    NLOHMANN_DEFINE_TYPE_EXTENSION(JsonGlyph, letter, x0, y0, dx, pixelWidth, pixelHeight, s0, t0, s1, t1);
+
+    class JsonFont
+    {
+    public:
+        int pixelHeight;
```
Laupetin: The game allows this to be negative technically but it never is in any vanilla fonts, so i'd make this `unsigned` as well, as a negative number does not make sense

src/ObjLoading/Font/FontLoader.cpp.template
```diff
@@ -0,0 +1,211 @@
...
+                font.glyphs = nullptr;
+                return true;
+            }
+
+            font.glyphs = m_memory.Alloc<Glyph>(font.glyphCount);
+            for (auto i = 0; i < font.glyphCount; i++)
+            {
+                if (!CreateGlyphFromJson(jFont.glyphs[i], font.glyphs[i], font))
+                    return false;
+            }
```
Laupetin, line 232: The game actually expects these glyphs to be sorted.
When just reading the dumped fonts, these sorting is already applied as the order is not changed in the array.

But it's still a good idea to properly sort the glyphs after loading in case of changes or custom fonts:

The first 95 glyphs must be ASCII codes 0x20 to 0x7F (inclusive), sorted ascending.
Then any following glyphs are optional and must be sorted ascending as well.

(See: `R_GetCharacterGlyph`)

src/ObjWriting/Font/FontDumper.cpp.template
```diff
@@ -0,0 +1,119 @@
...
+        {
+        }
+
+        void Dump(const Font_s& font) const
+        {
+            JsonFont jFont;
+            CreateJsonFont(jFont, font);
+
+            ordered_json jRoot;
+            jRoot["$schema"] = "http://openassettools.dev/schema/font.v1.json";
```
Laupetin, line 92: Todo: Add this schema in the docs repo

Laupetin: See https://github.com/OpenAssetTools/Docs/pull/127

Laupetin (approved):
Thank you very much @michaeloliverx but also @uhhashe, as I already integrated the string part from you here (and added you as coauthor).

Also thank you @MrIkso for working on the base PR previously, will also add you as a coauthor :D

MrIkso:
Good. This PR looks better than the one I previously created. Thanks

michaeloliverx:
Good teamwork folks!

PR #868, opened 2026-06-29
feat: add log file

uhhashe:
Adds a parameter to write console messages to a log file for easier viewing
Example:
image


Laupetin:
Thanks for the PR 😄 Generally there are other standard ways to achieve the same result, so is there a particular reason why this needs to be a built in feature rather than just making use of standard command line features?

e.g. you can redirect the output to a file like this with the standard features of your console:
`./Linker test_mod > ./log.txt`
or
`./Linker test_mod 2>&1 > ./log.txt` if you want to redirect both errors and normal log output

or if you want to see the output additionally to writing it to a file:

`./Linker test_mod | tee ./log.txt`
or
`./Linker test_mod 2>&1 | tee ./log.txt`
respectively.

Is there an advantage to providing a built-in way that you are missing otherwise?

uhhashe:
huh I somehow did not know that, that's very helpful to know thanks. About the overall idea, it was brought up to me to from a few people who wanted to log the console to a file, perhaps this could be printed in the `--help` command when run?

PR #869, opened 2026-06-29
feat: T4 weapon loading and dumping

michaeloliverx:


Laupetin (approved):
Okay you aced the weapon assets which i kind of was dodging because they are so jarring to do 😅 So thank you very much for taking that effort ❤️

PR #878, opened 2026-07-04
feat: T5 map ents dumper

uhhashe:
Adds map ents dumping support for T5
T5 is the only game without this


Laupetin (approved):
> T5 is the only game without this

At least the only game that does not make use of additional data (with iw4,iw5) 😄 Thanks

PR #879, opened 2026-07-04
feat: T4 and T5 physconstraints dumping

uhhashe:
Adds dumping support for `PhysConstraints` asset for T4 and T5


src/ObjCommon/Game/T4/PhysConstraints/PhysConstraintsFields.h
```diff
@@ -0,0 +1,86 @@
...
+        {"c4_shakescalex", offsetof(PhysConstraints, data[3].scale[1]),     CSPFT_FLOAT },
+        {"c4_shakescaley", offsetof(PhysConstraints, data[3].scale[2]),     CSPFT_FLOAT },
+        {"c4_shakescalez", offsetof(PhysConstraints, data[3].scale[3]),     CSPFT_FLOAT },
+        {"c4_min_angle",   offsetof(PhysConstraints, data[3].minAngle),     CSPFT_FLOAT },
+        {"c4_max_angle",   offsetof(PhysConstraints, data[3].maxAngle),     CSPFT_FLOAT },
+        {"c4_yaw",         offsetof(PhysConstraints, data[3].scale[1]),     CSPFT_FLOAT },
+        {"c4_pitch",       offsetof(PhysConstraints, data[3].scale[2]),     CSPFT_FLOAT },
+    };
+
+    inline const char* s_constraintTypeNames[]{"none", "point", "distance", "hinge", "joint", "actuator", "fake_shake", "launch", "rope"};
```
Laupetin: If you have such an array, best use a trailing comma to make clang-format give each entry its separate line

src/ObjCommon/Game/T5/PhysConstraints/PhysConstraintsFields.h
```diff
@@ -0,0 +1,97 @@
...
+        "point",
+        "distance",
+        "hinge",
+        "joint",
+        "actuator",
+        "fake_shake",
+        "launch",
+        "rope",
+        "light",
+    };
```
Laupetin, line 96: I know this is done like T6, it's probably a bit older there but since you brought up the location here: Checking the strings array with:

```cpp
    static_assert(std::extent_v<decltype(s_constraintTypeNames)> == NUM_CONSTRAINT_TYPES);
```

makes sure it is in sync with the enum and also documents which size this should have and why.

Laupetin (dismissed):
Thanks for giving this asset a bit of feature parity :D 
Also what a weird asset... vanilla does not even load it fully, only 4/16 constraints.

Also starting from constraint 2 the vanilla game is missing a bunch of data as well (that is still dumped here), for all 3 titles o.O
But since i apparently did that for T6 as well i guess it should be the same here
(Vanilla would usually be missing `c2_spin_scale`, `c2_shakescalex`, `c2_shakescaley`, `c2_shakescalez`, `c2_yaw`, `c2_pitch` and the same for c3 and c4)

Laupetin (approved):

PR #883, opened 2026-07-06
refactor: use standard library endian utilities

michaeloliverx:
- Replaces custom byte-swap/platform-endian code with `std::endian` and `std::byteswap`.
- Removes the now-empty `Endianness.cpp` implementation file.

## Notes
- `std::byteswap` requires C++23, which the project already uses.


Laupetin (approved):
looks good, thanks!

PR #884, opened 2026-07-06
feat: Add ANTLR4 Parsing Support and Replace PhysPreset Loading

pagingred:
This PR implements this following:
- Adds a git submodule and thirdparty premake entry for antlr4
  - Builds antlr4 static lib and links to headers
  - Requires ANTLR4CPP_STATIC be defined
  - Generates c++ parser, lexer, tokens, and interpreter file for each input .g4 grammar file
- When using `--antlr4` flag, assumes developer has Java in their PATH
  - Flag determines if we want to regenerate or not.
- Replaces PhysPreset loader for all games as a proof of concept
  - Added test cases for all games, all PASS
  - Additional logic for FlameTable, ZBarrier, Tracer, PhysConstraint, MenuDef, MenuList, and Glasses is ready for PRs after the base is verified.


src/ObjCommon/InfoString/InfoString.h
pagingred, line 1: Only did this because my Windows build failed to complete when the constexpr operator(...) function calls the non-constexpr function utils::StringEqualsIgnoreCase(...)

pagingred:
If desired I could put together an example / test. It could contain a baseline grammar and show how to "get the data out" of the parsed AST

EDIT: There are now tests for loading a PhysPreset for each game.

thirdparty/antlr4.lua
```diff
@@ -0,0 +1,138 @@
...
+
+    if result_str ~= "OK" then
+        premake.error("Failed to download ANTLR4 jar from " .. downloadUrl .. ": " .. result_str)
+    end
+
+    io.writefile(versionFile, targetVersion)
+    return jarPath
+end
+
+function antlr4:generateGrammar()
```
Laupetin: This is not used?

As it is not used, hard to comment on, but in general:

This must only be called on an optional flag set for premake5 that enables generation of parsers. Java should not be a hard requirement on being able to compile the project. All generated (code) output should be checked in into the repository.

thirdparty/antlr4.lua
```diff
@@ -0,0 +1,138 @@
...
+        os.execute(cmd)
+
+        prebuildcommands { cmd }
+
+        files {
+            path.join(outdir, "**.h"),
+            path.join(outdir, "**.cpp")
+        }
+
+        includedirs { outdir }
```
Laupetin: How is this supposed to be used? This just matches all grammars in the entire project folder but sets up files and include dirs?
This makes it so only one project could do this, while there may be multiple. It does also not respect the project folders at all.

thirdparty/antlr4.lua
```diff
@@ -0,0 +1,138 @@
...
+        local cmd = "java -jar \"" .. jar .. "\" -Dlanguage=Cpp -o \"" .. outdir .. "\" \"" .. path.getabsolute(grammar) .. "\""
+
+        print("  " .. rel)
+
+        if not os.isdir(outdir) then
+            os.mkdir(outdir)
+        end
+        os.execute(cmd)
+
+        prebuildcommands { cmd }
```
Laupetin: This is not setup as a custom build step with proper input and outputs.
This makes it so the grammar gets always generated, regardless of whether it was changed or not.
It always overwrites the source files so the compiler would always regenerate it, making it so the build is never fully doing something, no matter how many times you call it.

thirdparty/antlr4.lua
```diff
@@ -0,0 +1,138 @@
...
+        local rel = path.getrelative(folder, grammar)
+        local outdir = path.getabsolute(path.join(buildFolder, "src", path.getdirectory(rel)))
+        local cmd = "java -jar \"" .. jar .. "\" -Dlanguage=Cpp -o \"" .. outdir .. "\" \"" .. path.getabsolute(grammar) .. "\""
+
+        print("  " .. rel)
+
+        if not os.isdir(outdir) then
+            os.mkdir(outdir)
+        end
+        os.execute(cmd)
```
Laupetin: Why is this executed immediately?

thirdparty/antlr4.lua
Laupetin: As generation of grammars should not be required, having a CI step that checks whether the checked in source files are up to date would be very helpful 😅

thirdparty/antlr4.lua
```diff
@@ -0,0 +1,138 @@
...
+    local jar = path.getabsolute(self:installJar())
+
+    print("Running antlr4 generation...")
+
+    for _, grammar in ipairs(grammars) do
+        local rel = path.getrelative(folder, grammar)
+        local outdir = path.getabsolute(path.join(buildFolder, "src", path.getdirectory(rel)))
+        local cmd = "java -jar \"" .. jar .. "\" -Dlanguage=Cpp -o \"" .. outdir .. "\" \"" .. path.getabsolute(grammar) .. "\""
+
+        print("  " .. rel)
```
Laupetin: leftover debug code?

src/ObjCommon/InfoString/InfoString.h
Laupetin: That should not happen if you use the latest VS Studio version? Or what are you using to build?
It works on both CI and my dev environment, so there must be a difference in yours.

In general, yes, true. Would be worth considering making `utils::StringEqualsIgnoreCase` constexpr instead?
But generally, probably better fit in a separate PR 😅

thirdparty/antlr4.lua
```diff
@@ -0,0 +1,138 @@
...
+        defines { "ANTLR4CPP_STATIC" }
+
+        self:include(Includes:create())
+
+        warnings "off"
+end
+
+function antlr4:jarDir()
+    return path.join(BuildFolder(), "thirdparty/antlr4")
+end
```
Laupetin: I don't think this needs to be a separate method, right? You didn't make separate methods  for the other dirs as well

.gitmodules
```diff
@@ -22,3 +22,6 @@
 [submodule "thirdparty/webwindowed"]
 	path = thirdparty/webwindowed
 	url = https://github.com/Laupetin/webwindowed.git
+[submodule "thirdparty/antlr4"]
+	path = thirdparty/antlr4
+	url = https://github.com/antlr/antlr4.git
```
Laupetin, line 30: You added a submodule entry to the `.gitmodules` file, but you didn't add the submodule itself.
When I clone the branch, i do not have a `thirdparty/antlr4` folder because of that.

Submodules should only be added with the appropriate `git submodule add ...` command and not by adjusting this file manually 😅

thirdparty/antlr4.lua
Laupetin, line 1: Generally, adding ANTLR4 is a big task and a not so trivial one as well in addition to that.
First time introducing this dependency does require fundamental work beforehand.

The project already has a bunch of parser and preprocessor logic. Since it's not a long-term goal to have two separate parser systems, introducing ANTLR means setting a foundation to replace the old parser systems with ANTLR. To be able to do that it would be very nice to evaluate how and if the current systems can be brought to it in the current form.

That especially applies to the preprocessor stuff that i previously hand-tinkered that will still be needed for antlr.
Unless you have a good way to remake that in a way that is more appropriate for using ANTLR.
E.g. Handling comments in ANTLR is pretty easy as there is already samples on how to do this in the official antlr grammars repo.

So before I can merge this and bring ANTLR4 in the project, this PR should at the very least show how the following migrations from the hand-tinkered parsing code to ANTLR4 could work and look like (or if you wanna keep this initial PR small, there should at least be an additional PR that does it).
That is not a small or easy task and also why the issue for it has been lying around for a bit 😅

src/ObjCommon/InfoString/InfoString.h
kuromiii: I had the same issue on my system, using VS2026, couldn't build without this exact change.
EDIT: Updated to June 2026 Feature update and now seems to work...? But I definitely had the issue before updating.

pagingred: Created an issue #905 for this, updating my build tools for now.

thirdparty/antlr4.lua
pagingred: Changed to execute only when installing the jar.

thirdparty/antlr4.lua
pagingred: Changed to proper build step.

thirdparty/antlr4.lua
pagingred: Removed.

thirdparty/antlr4.lua
pagingred: Correct. However refactored code uses a separate method which also calls this, so it is now better as a separate method.

.gitmodules
pagingred: My bad here lol, I corrected this and removed the stale module before re-adding it properly.

thirdparty/antlr4.lua
pagingred: Changed to proper build step.

thirdparty/antlr4.lua
pagingred: Changed now so it outputs the generated files in the same directory as the source .g4 file.

thirdparty/antlr4.lua
pagingred: Added a --antlr4 flag. Only when enabled do we try to generate parser code with the available JAVA executable and the ANTLR4 jar.

thirdparty/antlr4.lua
pagingred: For a working proof of concept, I replaced all PhysPreset loader code with an ANTLR4 .g4 grammar file, its generated parser code, and game-specific implementations. 

I can continue to put loaders into this one PR if that is desired, but I'd imagine we want multiple PRs for those conversions. What are your thoughts?

Laupetin: I would not replace PhysPreset parsing or any other infostring based asset with ANTLR4, that's not what I mean 😅

Infostrings are a very simple format and also all work the same. So _if_ there was an antlr grammar for it, it would be one for Infostrings generally. But i wouldnt do that. Because that's very overkill. You dont need a lexer + parser for that.

What I meant that needed a proof of concept is the things that are currently using the Parser component that does something very similar to antlr4 with a lexer + parser layer, a parser state etc.
Infostrings dont do that. They just read a string and separate it at backslashes. 

For example zone file parsing.

I'd not replace PhysPreset parsing with antlr even when antlr was already in the project tbh.

pagingred:
Turns out the request was to replace the src/Parser with ANTLR, not individual parsers. So for now this PR is not required.

PR #894, opened 2026-07-10
feat: QOS (007: Quantum of Solace) support

### LINKED ISSUE #349: Feature Request: Support for Quantum of Solace (2008)
wondering if support for this game is on your roadmap. It uses the IW3 engine

michaeloliverx:
image

image

## Summary

Adds initial support for `QOS` / `007: Quantum of Solace` PC fastfiles.

Closes https://github.com/Laupetin/OpenAssetTools/issues/349.

The current goal is not to claim that every QoS struct is fully understood. It is to make the retail PC fastfiles loadable, unlinkable, and inspectable while keeping the inferred parts of the layouts obvious for future work.

## Notes

QoS is close to the IW3/T4 engine family. The structs in this PR are mostly inferred from the QoS PC and Wii loader/decomp, with IW3/T4 used as a guide where the code paths clearly line up.

Some details that may be useful while reviewing:

- The QoS fastfile header does not seem to use the usual ASCII magic string. The first header word is the little-endian zone version, and the following header words are used for xblock sizes.
- The xfile blocks found in the executable are `temp`, `runtime`, `virtual`, `large`, and `physical`.
- The QoS asset enum comes from the executable asset-name table.
- A Wii decomp with some symbol names was useful for function cross-reference, but struct sizes/offsets were not copied from Wii.
- The executable still has IWD support, but the retail PC image data checked here is bundled in fastfiles.
- Embedded QoS image data needed smallest-to-largest mip handling, matching IWI-style ordering.
- QoS xmodels needed some QoS-specific handling around packed vertex/material data rather than using the IW3/T4 assumptions directly.

Unknown fields are intentionally left as unknown/padding rather than given confident names too early. The current names can be tightened as more QoS-specific code paths are reversed.

## Current coverage

Added a few easy wins for dumpers and loaders and fleshed out the zone writing as well.

QoS is also wired into ModMan; images and xmodels can be inspected there.

## Validation
- Only tested on Windows
- Unlinked all 57 fastfiles from the English version successfully.
- Verified the final unlink output had all expected zone folders and generated `.zone` files.
- Linked a small QoS test zone using disk-loaded rawfiles and `--load`-referenced existing assets, then listed/unlinked the result as a basic roundtrip check.
- Tested ModMan inspection with QoS zones and confirmed images/xmodels render.

---

One known dump-time filename collision can happen with `ui_mp.ff`:

```text
ERROR: Failed to open file 'ui_mp'
Dumped rawfile "ui_mp"
Unloaded zone "ui_mp"
```

This happens because the dump output can contain a nested `ui_mp` directory and a trailing rawfile asset also named `ui_mp`.

I guess this can happen with any game in OAT.

## Useful Resources
- https://github.com/ProjectConsolation
- https://github.com/ProjectConsolation/qos-xport
- https://github.com/Scobalula/Greyhound/blob/22c3d17f7428573042b56f43af0ed2585989bebd/src/WraithXCOD/WraithXCOD/GameQuantumSolace.cpp


src/ObjWriting/XModel/XModelToCommonConverter.cpp.template
```diff
@@ -70,6 +73,30 @@ namespace
         if (potentialTextureDefs.size() == 1)
             return GetImageFromTextureDef(*potentialTextureDefs[0]);
 
+#ifdef FEATURE_QOS
```
michaeloliverx, line 76: This feels hacky, you can probably find a better solution than me

src/ZoneLoading/Game/QOS/ZoneLoaderFactoryQOS.cpp
```diff
@@ -0,0 +1,142 @@
+#include "ZoneLoaderFactoryQOS.h"
```
michaeloliverx, line 1: QoS seems to be the odd one out here because its fastfile header does not have the usual ASCII magic string. The first 4 bytes of `ZoneHeader::m_magic` are the zone version (`470`), while `ZoneHeader::m_version` is actually the first xblock size. The remaining xblock sizes follow directly after the normal header.

That is why this factory does not use the usual `StepLoadZoneSizes`/`StepAllocXBlocks` sequence and has a small QoS-specific allocation step instead.

Feel free to refactor it

src/Common/Game/QOS/QOS.h
```diff
@@ -0,0 +1,123 @@
...
+    };
+
+    struct XAssetList
+    {
+        ScriptStringList stringList;
+        int assetCount;
+        XAsset* assets;
+    };
+
+    using AssetXModelPieces = Asset<ASSET_TYPE_XMODELPIECES, XModelPieces>;
```
michaeloliverx: Sorry I did not clean these up, I will leave that to you 😛

src/Common/Game/QOS/GameQOS.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+#include "QOS.h"
+
+#include <type_traits>
+
+using namespace QOS;
+
+namespace
+{
+    constexpr const char* ASSET_TYPE_NAMES[]{
```
michaeloliverx, line 11: Sorry I forgot to normalise these

ujicos:
Tested and confirmed working on QoS PC version 1.1, via @ProjectConsolation .
Generated a test fastfile "patch_consolation.ff", which successfully overrode dm.gsc from zone.
Thank you for your work, Michael!
image

src/Common/Game/QOS/CommonAssetQOS.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+#include <cassert>
+
+namespace QOS
+{
+    CommonAssetTypeMapper::CommonAssetTypeMapper() = default;
+
+    CommonAssetType CommonAssetTypeMapper::GameToCommonAssetType(const asset_type_t gameAssetType) const
+    {
+        static CommonAssetType lookupTable[static_cast<unsigned>(ASSET_TYPE_COUNT)]{
```
Laupetin: I know this is based on my previous mistake but i'm trying to fix this whenever i see it and i just noticed it here: This does not work as you might expect unfortunately 😅 

I was originally thinking this would check whether the array is actually `ASSET_TYPE_COUNT` entries long and complain if it isn't, but it is not. It will only complain if you add _too many_ entries in the initializer list, but not if you add _too few_ (it will add default initialized entries too fill up to the specified size in that case) :/

So i'm replacing this everywhere with no count between the array brackets and a static_assert instead:
```cpp
        static_assert(std::extent_v<decltype(lookupTable)> == static_cast<unsigned>(ASSET_TYPE_COUNT));
```

src/ObjCommon/XModel/JsonXModel.h.template
```diff
@@ -16,6 +16,8 @@
 #define FEATURE_T5
 #elif GAME == "T6"
 #define FEATURE_T6
+#elif GAME == "QOS"
+#define FEATURE_QOS
```
Laupetin: to keep things clear i'd say it's probably best to keep the conditions alphabetically sorted by game name?

src/ObjImage/Image/Dx9TextureLoader.cpp
```diff
@@ -90,16 +97,27 @@ namespace image
         texture->Allocate();
         const auto mipMapCount = m_has_mip_maps ? texture->GetMipMapCount() : 1;
         const auto faceCount = m_type == TextureType::T_CUBE ? 6 : 1;
-        const void* currentDataOffset = data;
+        auto* currentDataOffset = static_cast<const uint8_t*>(data);
 
-        for (auto currentMipLevel = 0; currentMipLevel < mipMapCount; currentMipLevel++)
+        const auto copyMipLevel = [&](const int currentMipLevel)
```
Laupetin: I usually try to avoid lambdas as i fear they create a bit of unnecessary overhead.
In this case this looks like a place the rest of the codebase would just use a normal function, no?

src/ObjWriting/Game/QOS/RawFile/RawFileDumperQOS.cpp
Laupetin, line 1: This can be templated, no? There's already a template for MapEnts dumping

src/ObjWriting/XModel/XModelToCommonConverter.cpp.template
```diff
@@ -70,6 +73,30 @@ namespace
...
+            auto name = std::string_view(image->name);
+            while (!name.empty() && (name.front() == ',' || name.front() == '*'))
+                name.remove_prefix(1);
+
+            if (name.ends_with("_c") || name.ends_with("_col") || name.ends_with("_color") || name.ends_with("_diff"))
+                return image;
+
+            if (!qosFallback && !name.ends_with("_edge") && !name.ends_with("_print") && !name.ends_with("_prsz")
+                && name.find("_edge_") == std::string_view::npos)
+                qosFallback = image;
```
Laupetin, line 95: nit and personal taste: for multi line if conditions i'd always use curly braces and not omit them for clearity

src/ObjWriting/XModel/XModelToCommonConverter.cpp.template
Laupetin: hm i mean this is not too important and doesn't have a 100% correct answer too be honest.
This is just a way to have something to look at when looking at the 3D model but is not used when loading the data again so if it is somehow helpful when presenting the model and looks okayish, sure why not.

If it produces better results like this, i don't mind having it that way. Though i'd restructure it a bit as soon as other games require more complex handling like this as well.

src/ObjWriting/XModel/XModelToCommonConverter.cpp.template
```diff
@@ -346,8 +376,17 @@ namespace
                 vertex.coordinates[1] = v.xyz.y;
                 vertex.coordinates[2] = v.xyz.z;
                 Common::Vec3UnpackUnitVec(v.normal, vertex.normal);
+#ifdef FEATURE_QOS
+                vertex.color[0] = 1.0f;
+                vertex.color[1] = 1.0f;
+                vertex.color[2] = 1.0f;
+                vertex.color[3] = 1.0f;
```
Laupetin, line 385: What's the reason for always setting full white here? The vertex does have a color property, no?

src/ObjWriting/Game/QOS/RawFile/RawFileDumperQOS.cpp
michaeloliverx: This PR predated the templated dumper I added

src/ZoneLoading/Game/QOS/ZoneLoaderFactoryQOS.cpp
Laupetin: Added a way to peek at the first few bytes of the file in a generic way that does not assume the game uses the classic ZoneHeader.

Since the game's headers and therefore peek sizes are different, the ZoneLoader does not start after the zone header anymore but at the start of the file.
This means that any ZoneLoaderFactory must add a step to skip the respective header to have the same behaviour as before.

src/Common/Game/QOS/QOS_Assets.h
```diff
@@ -0,0 +1,2535 @@
...
+            float r;
+            float g;
+            float b;
+            float a;
+        };
+    };
+
+    typedef unsigned short ScriptString;
+    typedef unsigned char raw_byte;
+    typedef unsigned int raw_uint;
```
Laupetin: These typedefs unfortunately cause a problem 😅 OAT marks pointers for reusage when at least one usage of the type is reusable because that's a requirement of the games:
When there are two usages of a type but only one is reusable, the reusable one might reuse a pointer from the non-reusable usage.

OAT makes sure even non-reusable pointers are marked based on the type... and the typedef here is enough to satisfy this constraint. There is at least one reusable usage of `raw_uint`, so all `raw_uint` usages are marked for potential reuse there.

The issue arises when this happens on members that are inside a runtime block, because those are not actually written to the zone so an assertion triggers.

As most simple measure to fix this and to not mark pointers unnecessarly i'll just remove those two typedefs and replace them with `unsigned char`, `unsigned int` respectively, sorry 😅

src/ZoneWriting/Game/QOS/ContentWriterQOS.cpp
```diff
@@ -0,0 +1,211 @@
...
+        WRITE_ASSET(ASSET_TYPE_PHYSCONSTRAINTS, PhysConstraints, physConstraints)
+        WRITE_ASSET(ASSET_TYPE_DESTRUCTIBLEDEF, DestructibleDef, destructibleDef)
+        WRITE_ASSET(ASSET_TYPE_XANIMPARTS, XAnimParts, parts)
+        WRITE_ASSET(ASSET_TYPE_XMODEL, XModel, model)
+        WRITE_ASSET(ASSET_TYPE_MATERIAL, Material, material)
+        WRITE_ASSET(ASSET_TYPE_TECHNIQUE_SET, MaterialTechniqueSet, techniqueSet)
+        WRITE_ASSET(ASSET_TYPE_IMAGE, GfxImage, image)
+        WRITE_ASSET(ASSET_TYPE_SOUND, snd_alias_list_t, sound)
+        WRITE_ASSET(ASSET_TYPE_SND_CURVE, SndCurve, sndCurve)
+        WRITE_ASSET(ASSET_TYPE_CLIPMAP_SP, clipMap_t, clipMap)
```
Laupetin, line 105: This is missing clipmap mp 😅 

I found this when running the [manual "rebuild all zones" system test](https://github.com/Laupetin/OpenAssetTools/blob/main/test/SystemTests/RebuildVanillaZonesTests.cpp), can only recommend this for a quick and trivial check whether rebuilding all zones via reusing their assets from in-memory is at least successful:
```shell
OAT_FASTFILE_DIR=C:/Cod/Qos/zone ./build/lib/Debug_x86/tests/SystemTests.exe 'Rebuild vanilla zones'
```

michaeloliverx: I did not know of this test... I have been doing it with a one off script each time

src/ObjWriting/XModel/XModelToCommonConverter.cpp.template
michaeloliverx: I can't remember the exact model now, but unpacking the QOS vertex color was giving weird tinting/artifacts in the ModMan model viewer. That prompted me to compare against Greyhound.

Greyhound does define a QOS vertex buffer with `Color[4]`, but in the Quantum Solace branch it ignores `VertexInfo.Color` and hardcodes RGBA to `0xFF`:
https://github.com/Scobalula/Greyhound/blob/22c3d17f7428573042b56f43af0ed2585989bebd/src/WraithXCOD/WraithXCOD/CoDXModelTranslator.cpp#L440-L465

Seemed to fix the issue I was having.

src/Common/Game/QOS/QOS_Assets.h
```diff
@@ -0,0 +1,2648 @@
...
+
+    struct GfxPackedVertex
+    {
+        vec3_t xyz;
+        float binormalSign;
+        GfxColor color;
+        // This is one of those values:
+        // 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 14, 15, 17
+        // The vertexdecl seems to ignore this
+        unsigned int unknown0; // Offset 20; Greyhound treats this as padding.
```
Laupetin, line 391: I tried to find out more about this field for a while but was only mildly successful.
It does not seem to be used with the vertexdecl so cannot be used within shaders.

Not sure if this is even actually used outside of shaders, would have to debug the game for it.
But very odd field for sure. Will leave it at that for now, but added a bit of information i gather while researching this in the PR.

src/ObjWriting/XModel/XModelToCommonConverter.cpp.template
Laupetin: This seems to be the artifacts then?



Yeah idk, i guess we can keep it like that for now. idk if this might become a problem when trying to _write_ QOS models but we will see

Laupetin (approved):
very nice, thank you very much 😄 
For now i'll keep it like this and not invest more time into researching the struct details for the initial PR.
That can be done for future asset support then.
Though I gotta say the game has more differences and changes compared to the other titles (like IW3, T4, T5) than i would have thought.
Asset support as full as the other titles may not be entirely feasible but the future will tell.
It's a lot of work at the very least.

PR #904, opened 2026-07-14
feat: log detected game when loading zones

michaeloliverx:


src/Unlinking/Unlinker.cpp
```diff
@@ -81,6 +81,14 @@ namespace
         return AutoSearchPaths::GetForGame(gameId)->GetSearchPathsForZonePath(zonePath);
     }
 
+    void LogLoadedZone(const Zone& zone)
+    {
+        const auto* game = IGame::GetGameById(zone.m_game_id);
+
+        con::info("Loaded zone \"{}\"", zone.m_name);
+        con::info("Detected game: {} ({})", game->GetFullName(), game->GetShortName());
```
Laupetin: What do you think about

```suggestion
        con::info("Loaded zone \"{}\" ({})", zone.m_name, game->GetShortName());
```

because i feel like right now these two messages do not seem entirely connected.
Also makes it a bit less verbose.

Laupetin: I feel like when adjusting it for Unlinker, it might be worth adjusting in Linker as well?
https://github.com/Laupetin/OpenAssetTools/blob/0c1efa1667d47ab4e22d38fd2d97b73d4a383543/src/Linking/Linker.cpp#L413

michaeloliverx: Makes sense. Forgot to put in the desc the motivation for this is OAT doesn't stop you from `--load`ing fastfiles from other games. So it can be useful as an audit trail

PR #907, opened 2026-07-15
fix: IW3 ensure `loadForRenderer` is set for shaders

### LINKED ISSUE #770: [Linker][IW3] Relinking a map results in broken in game rendering
Relinking an IW3 fastfile using the `--load` mechanism results in a broken map in terms of rendering. Reproduced using both x86 and x64 builds.
Image

michaeloliverx:
This was a fun one 🙃 

Stock `mp_showdown` used the following values:

| Shader profile   | `loadForRenderer` |
| ---------------- | ----------------: |
| Shader Model 1.x |                 0 |
| Shader Model 2.0 |                 0 |
| Shader Model 3.0 |                 1 |

The sky is still not rendering but at least the world is.

Screenshot 2026-07-15 195626

---

**EDIT**

I was using bad input fastfile (again 😢) so the statement above is not true.. BUT thankfully setting all shaders `loadForRenderer` to `1` fixes the issue completely!

Fixes https://github.com/Laupetin/OpenAssetTools/issues/770

### DIFF
--- a/src/ObjCompiling/Techset/TechsetCompiler.cpp.template
+++ b/src/ObjCompiling/Techset/TechsetCompiler.cpp.template
@@ -69,6 +69,28 @@ namespace
         return static_cast<MaterialWorldVertexFormat>(0);
     }
 
+#if defined(FEATURE_IW3)
+    uint16_t GetLoadForRenderer(const std::string& techniqueSetName)
+    {
+        if (techniqueSetName.starts_with("sm2/"))
+            return 0;
+        return 1;
+    }
+
+    void ApplyLoadForRendererToTechnique(const MaterialTechnique& technique, const uint16_t loadForRenderer)
+    {
+        for (auto passIndex = 0u; passIndex < technique.passCount; passIndex++)
+        {
+            auto* vertexShader = technique.passArray[passIndex].vertexShader;
+            if (vertexShader)
+                vertexShader->prog.loadDef.loadForRenderer = std::max(vertexShader->prog.loadDef.loadForRenderer, loadForRenderer);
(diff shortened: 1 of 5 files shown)


Laupetin:
Not sure why you think that your first theory was wrong 😅 Looking through the game's code this seems entirely correct to me. Besides the fact maybe that there is no shader model 1 in use.

IDA shows me weird output for IW3 but on IW4 you can clearly see this is what the game does:
```cpp
void Load_CreateMaterialVertexShader(GfxVertexShaderLoadDef *loadDef, MaterialVertexShader *vertexShader)
{
  if ( loadDef->program
    && r_loadForRenderer->current.enabled
    && loadDef->loadForRenderer == r_rendererInUse->current.integer )
  {
    // Load the shader
  }
  else
  {
    vertexShader->prog.vs = 0;
  }
}
```

The `r_rendererInUse` dvar is an enum dvar with the following values:
* 0 = "Shader model 2.0"
* 1 = "Shader model 3.0"
* 2 = "Default"

Funnily enough, IW4 and IW5 also have the dvar `r_rendererInUse` that is checked, but it has only 2 entries: SM3 and default.

Edit: Ignore me, there definitely is sm1 etc in use and idk whats going on

Laupetin:
Okay I adapted the IW3 Modtools Linker's logic for determining the loadForRenderer value and it seems to work perfectly when testing with mp_showdown.

It sets the loadForRenderer value of all shaders in techsets with `sm2/` prefix in their names to `0` and otherwise to `1`.
The code now takes the max to make sure any shader that is used in an sm3 techset has its loadForRenderer value set to `1`.

Laupetin (approved):
thank you very much for tracking down the issue :D

PR #908, opened 2026-07-15
fix: prevent IWD paths from matching nested filenames

### LINKED ISSUE #887: IWD user paths include files that it shouldn't
If you setup your mod like this:
zone_raw/mod/mod.zone
```
>iwd,mod
```

and create `zone_raw/mod/iwd/maps/mp/mod.txt`

it will create this in the iwd:
```
   Date      Time    Attr         Size   Compressed  Name
------------------- ----- ------------ ------------  ------------------------
2026-07-07 20:05:40 .....            4            6  ..\maps\mp\mod.txt
------------------- ----- ------------ ------------  ------------------------
2026-07-07 20:05:40                  4            6  1 files
```

This is caused by some whacky code in `SearchPathFilesystem.cpp` that incorrectly matches files that it shouldn't because it incorrectly modifies the prefix.

Thank you @LJW-Dev for discovering this!

michaeloliverx:
Fixes https://github.com/Laupetin/OpenAssetTools/issues/887

### DIFF
--- a/src/ObjCommon/SearchPath/SearchPathFilesystem.cpp
+++ b/src/ObjCommon/SearchPath/SearchPathFilesystem.cpp
@@ -31,7 +31,7 @@ namespace
                 return false;
 
             path = parentDir.string();
-            prefix = combinedPath.filename().string();
+            prefix = combinedPath.string();
         }
 
         return true;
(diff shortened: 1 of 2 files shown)


Laupetin (approved):
Works great, thank you very much 🎉

Issue #626, opened 2025-12-25
Feature parity for games


Laupetin:
To avoid confusion for users, supported games should at least have feature parity for assets, that all of them support.
To achieve that, more loading and dumping code should be either shared or generated to allow for minimal code duplication.

The following assets are currently implemented for one game but not others:
- [ ] PhysPreset
  - Supported: IW4, T6
  - Unsupported: IW3, IW5, T5
- [ ] (Addon)MapEnts
  - Dumps: IW3, T6
  - Unsupported: IW4, IW5, T5
- [ ] Menus
  - Supported: IW4, IW5
  - Unsupported: IW3, T5, T6
- [ ] Weapons
  - Supported: IW4, IW5, T6
  - Unsupported: IW3, T5
- [x] Image
  - Loads DDS: IW3
  - Loads IWI: IW5, T6
  - Unsupported: IW4, T5
- [x] Techset
  - Supported: IW3, IW4, T5, T6
  - Unsupported: IW5
- [ ] Tracer
  - Supported: IW4 (Only dump), T6
  - Unsupported: IW5
- [ ] Vehicle
  - Supported: IW4 (Only dump), T6
  - Unsupported: IW5

PR #909, opened 2026-07-16
feat: IW3 menu dumping and loading

michaeloliverx:
## Summary

Adds IW3 menu and menulist dumping. The original source cannot be reproduced exactly because macros and preprocessing are lost, but the output preserves runtime behaviour and can be relinked with the official tools.

## Verification

- Dumped stock `ui_mp.ff`
- Copied the generated menus and menulist into CoD4 Mod Tools
- Rebuilt a `mod.ff` using the official `linker_pc`
- Tested the relinked menus in game

## Notes

Loading every `ui_mp` menu in one mod exceeds stock CoD4’s 128-menufile limit:

```text
Exceeded limit of 128 'menufile' assets.
```

CoD4x can raise the limit for testing:

```powershell
.\iw3mp.exe +set r_xassetnum "menufile=512"
```

The stock linker also cannot load four generated menu files because their paths exceed its 63-character path limit. It reports them as missing even though the files exist:

```
Couldn't find menu file 'ui_mp/demolitions_popup_cac_defaultclasswarning_heavy_gunner.menu'
Couldn't find menu file 'ui_mp/demolitions_popup_cac_defaultclasswarning_demolitions.menu'
Couldn't find menu file 'ui_mp/heavygunner_popup_cac_defaultclasswarning_heavy_gunner.menu'
Couldn't find menu file 'ui_mp/heavygunner_popup_cac_defaultclasswarning_demolitions.menu'
link...compress...save...done.
```

The linker still completes and writes the fastfile.

## References

- https://github.com/promod/CoD4-Mod-Tools/blob/d9a5b2f17efe11fc33d4bfb9704972e9a7879e60/raw/ui/menudefinition.h
- https://github.com/SwagSoftware/KisakCOD/blob/ba3c79f3a5107aa30379178ce2c106081f29320b/src/ui/ui_shared.cpp

---

**EDIT**
### Update: IW3 menu loading

I have also added IW3 `.menu` loading in 57a0f6e91e53da343caa885a6cad6a82aba5ed16, since it was feasible to build on the existing shared menu parser used by IW4 and IW5.

The full round trip was tested by dumping every menu from the stock IW3 `ui_mp.ff` with OAT, linking the dumped menus back from disk into a new `mod.ff` using OAT, and verifying the result in game.


test/ObjWritingTests/Game/IW3/Menu/MenuDumperIW3Test.cpp
```diff
@@ -0,0 +1,360 @@
...
+        }
+        itemDef
+        {
+            name                        "button_test"
+            text                        "@MENU_TEST"
+            group                       "buttons"
+            rect                        10 20 180 24 1 2
+            style                       1
+            type                        1
+            visible                     when((localvarint("ui_highlight")==5&&localvarstring("ui_choicegroup")=="popmenu"));
```
michaeloliverx: The expressions could be printed better but I didn't investigate what it would take.

src/ObjCommon/Game/IW3/MenuConstantsIW3.h
```diff
@@ -0,0 +1,92 @@
+#pragma once
+
+#include "Game/IW3/IW3.h"
+
+#include <array>
+
+namespace IW3
+{
+    inline constexpr std::array<const char*, NUM_OPERATORS> g_expFunctionNames{
```
Laupetin: This kind of breaks with the other places where it uses a native array.
I honestly don't really know what the advantage of `std::array` would be here?

Also this has the same problem as the native array with a specified size:
You can accidentally add too little entries and it will fill up the array with default initialized entries which is suboptimal.
See here: https://github.com/Laupetin/OpenAssetTools/pull/894#discussion_r3700247743

This problem is actually not even solvable with `std::array`? Since you always have to specify a size?

src/ObjWriting/Menu/AbstractMenuWriter.cpp
```diff
@@ -117,6 +118,11 @@ namespace menu
         if (token.empty())
             return true;
 
+        char* numericEnd;
+        std::strtof(token.c_str(), &numericEnd);
+        if (numericEnd == token.c_str() + token.size())
+            return false;
```
Laupetin, line 127: The commit this happens in explains well why this is a thing, for later reference: A comment maybe?

src/ObjWriting/Game/IW3/Menu/MenuDumperIW3.cpp
```diff
@@ -0,0 +1,46 @@
...
+    std::string GetPathForMenu(menu::MenuDumpingZoneState* zoneState, const XAssetInfo<menuDef_t>& asset)
+    {
+        const auto menuDumpingState = zoneState->m_menu_dumping_state_map.find(asset.Asset());
+
+        if (menuDumpingState != zoneState->m_menu_dumping_state_map.end())
+            return menuDumpingState->second.m_path;
+
+        const auto* menuName = asset.Asset()->window.name;
+        const std::string fallbackName = menuName ? menuName : asset.m_name;
+        return "ui_mp/" + fallbackName + ".menu";
```
Laupetin: in which cases is `asset.Asset()->window.name` and `asset.m_name` not the same value? They should always be identical?

src/ObjCommon/Game/IW3/MenuConstantsIW3.h
michaeloliverx: My original motivation for using `std::array` was to make sure the number of names stays in sync with `NUM_OPERATORS`.

Other than sticking with the existing convention, is there any reason not to do something like:

```cpp
inline constexpr std::array g_expFunctionNames{
    ...
};

static_assert(g_expFunctionNames.size() == NUM_OPERATORS);
```

That way the size is inferred from the entries, and we still get the compile-time check.

I'm still quite early in my C++ journey, so I don't really have the intuition yet for when the older C-style constructs are actually preferable. I think I have a bit of an underlying bias that the more modern C++ alternatives are generally "better", so I'm interested to know if there's a practical downside here beyond consistency with the existing code.

Laupetin: > Other than sticking with the existing convention, is there any reason not to do something like:
>
> ```cpp
> inline constexpr std::array g_expFunctionNames{
>     ...
> };
> 
> static_assert(g_expFunctionNames.size() == NUM_OPERATORS);
> ```
> 
> That way the size is inferred from the entries, and we still get the compile-time check.

That would also would, yea. Didn't test with without template template arguments, it's just that you cannot specify it with just one.
It would work like this for `const char*` because the compiler can deduce that from your literals.
It wouldn't work for any struct type though as that one has to be specified and not deduced, for example: https://github.com/Laupetin/OpenAssetTools/blob/08631dcc76a7f869efb709f6c6d27d1ce5f037c6/src/ObjCommon/Game/IW4/Techset/TechsetConstantsIW4.h#L116-L170

So it's just more complicated to use, i don't think it would work in every circumstance (given it would work here), and in the end it is also just a wrapper around this:
```
_EXPORT_STD template <class _Ty, size_t _Size>
  class array {
  public:
  // [..]
  _Ty _Elems[_Size];
  // [..]
};
```

src/ObjLoading/Game/IW3/Menu/MenuConverterIW3.cpp
```diff
@@ -0,0 +1,645 @@
...
+    constexpr auto WINDOW_FLAG_NON_DEFAULT_FORECOLOR = 0x00010000;
+    constexpr auto WINDOW_FLAG_DECORATION = 0x00100000;
+    constexpr auto WINDOW_FLAG_HORIZONTAL_SCROLL = 0x00200000;
+    constexpr auto WINDOW_FLAG_AUTO_WRAPPED = 0x00800000;
+    constexpr auto WINDOW_FLAG_POPUP = 0x01000000;
+    constexpr auto WINDOW_FLAG_OUT_OF_BOUNDS_CLICK = 0x02000000;
+    constexpr auto WINDOW_FLAG_LEGACY_SPLIT_SCREEN_SCALE = 0x04000000;
+    constexpr auto WINDOW_FLAG_HIDDEN_DURING_FLASH_BANG = 0x10000000;
+    constexpr auto WINDOW_FLAG_HIDDEN_DURING_SCOPE = 0x20000000;
+    constexpr auto WINDOW_FLAG_HIDDEN_DURING_UI = 0x40000000;
```
Laupetin: Those are part of IW3_Assets.h, why are those duplicated here 😅

michaeloliverx: Oops 🙊

src/ObjLoading/Parsing/Menu/Sequence/ItemScopeSequences.cpp
```diff
@@ -130,8 +131,15 @@ namespace
                 throw ParsingException(pos, "Item must have be listbox to use this declaration");
         }
 
-        static void EnsureHasEditFieldFeatures(const CommonItemDef& item, const TokenPos& pos)
+        static void EnsureHasEditFieldFeatures(CommonItemDef& item, const TokenPos& pos)
         {
+            // ITEM_TYPE_TEXT is the default and can therefore be omitted from menu source.
+            if (item.m_type == 0 && item.m_feature_type == CommonItemFeatureType::NONE)
```
Laupetin: This should not check against a magic value zero but instead check whether the current item type supports the feature type `CommonItemFeatureType::EDIT_FIELD`.

That is:
- More explicit on what it is doing
- Does not rely on text being the default type for all games
- Does not rely on the default being 0

src/ObjLoading/Parsing/Menu/Domain/CommonItemDef.h
```diff
@@ -93,6 +93,7 @@ namespace menu
 
         std::string m_name;
         std::string m_text;
+        bool m_has_text = false;
```
Laupetin: `std::optional<std::string>`? That's the standard way for it if you want to differentiate between empty text and no text 😛

src/ObjLoading/Parsing/Menu/Matcher/MenuExpressionMatchers.cpp
```diff
@@ -50,6 +52,24 @@ std::unique_ptr<SimpleExpressionMatchers::matcher_t> MenuExpressionMatchers::Par
 
 const std::map<std::string, size_t>& MenuExpressionMatchers::GetBaseFunctionMapForFeatureLevel(const FeatureLevel featureLevel)
 {
+    if (featureLevel == FeatureLevel::IW3)
+    {
+        static std::map<std::string, size_t> iw3FunctionMap;
```
Laupetin: not that this is your mistake, but now that i see it: could be an `unordered_map` as no order is needed.

src/ObjLoading/Parsing/Menu/Sequence/ItemScopeSequences.cpp
```diff
@@ -91,6 +91,7 @@ namespace
 
             switch (featureLevel)
             {
+            case FeatureLevel::IW3:
             case FeatureLevel::IW4:
                 if (static_cast<unsigned>(type) >= std::extent_v<decltype(IW4_FEATURE_TYPE_BY_TYPE)>)
```
Laupetin: please make sure the name of this reflects that iw3 is also using it 😅

src/ObjLoading/Parsing/Menu/Sequence/ItemScopeSequences.cpp
```diff
@@ -491,7 +499,7 @@ namespace
             size_t baseColumnCount;
             const char* syntax;
             bool hasHeightValues;
-            if (state->m_feature_level == FeatureLevel::IW4)
+            if (state->m_feature_level == FeatureLevel::IW3 || state->m_feature_level == FeatureLevel::IW4)
             {
                 baseColumnCount = BASE_COLUMN_COUNT_IW4;
```
Laupetin: Please also make sure here that the constant states that iw3 uses it and not only iw4

src/ObjLoading/Game/IW3/Menu/LoaderMenuListIW3.cpp
```diff
@@ -0,0 +1,179 @@
...
+            : m_memory(memory),
+              m_search_path(searchPath)
+        {
+        }
+
+        AssetCreationResult CreateAsset(const std::string& assetName, AssetCreationContext& context) override
+        {
+            std::vector<menuDef_t*> menus;
+            AssetRegistration<AssetMenuList> registration(assetName);
+            auto& parsingState = context.GetZoneAssetCreationState<menu::MenuAssetZoneState>();
```
Laupetin: This is not the parsing state, the term "parsing state" is already used for storing the current state of the parser while traversing a single file.

This instead is the `zoneState` where it states state about already parsed menus over multiple assets.

Making a difference in name here makes this less confusing.

src/ObjLoading/Game/IW3/Menu/LoaderMenuListIW3.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+            std::vector<XAssetInfo<menuDef_t>*> menusOfFile;
+            menusOfFile.reserve(result.m_menus.size());
+            for (auto& commonMenu : result.m_menus)
+            {
+                auto converter = IMenuConverter::Create(ObjLoading::Configuration.MenuNoOptimization, m_search_path, m_memory, context);
+                auto* menuAsset = m_memory.Alloc<menuDef_t>();
+                AssetRegistration<AssetMenu> menuRegistration(commonMenu->m_name, menuAsset);
+                if (!converter->ConvertMenu(*commonMenu, *menuAsset, menuRegistration))
+                    return false;
```
Laupetin, line 158: IW4 has an error message here, IW3 does not: `con::error("Failed to convert menu file \"{}\"", commonMenu->m_name);`

src/ObjLoading/Game/IW3/Menu/LoaderMenuListIW3.cpp
Laupetin, line 1: One issue that i have with this (and a few other files as well) is that the menu code is currently not templated but duplicated and follows a very similar structure in both iw4 and iw5.

This implementation here is _prettyyyy_ similar but differs in method structure, variable names, reverses conditions and a few other tweaks here and there (besides the missing function defs).

Which is fine, I guess in some aspects this new structure is probably better, but this makes it very annoying to compare to the other game's when making a change to how fundamentally menus work, checking menu parsing implementation etc.

So maybe please if you wanna do some restructuring of the menu code, make sure that the IW4 and IW5 versions are aligned 😅 
So whenever possible, the code is structured the same when it is supposed to do the same (just for a different game).

test/ObjWritingTests/Game/IW3/Menu/MenuDumperIW3Test.cpp
Laupetin: i reused the code from iw4 now for a large part and it should print similar to there now.
Maybe you could test again, if the game still accepts these menus as expressions as always a bit tricky 😅 Especially since the way they work in menus is a bit odd in these games.

src/ObjLoading/Game/IW3/Menu/LoaderMenuListIW3.cpp
Laupetin: Restructured the functionality to be as identical as possible.
As a lot of things from this PR were also code improvements, i'll create a followup PR to bring these improvements to IW4 and IW5, so just from this PR alone, they won't be as identical as this comment might initially suggest. But they will be.

test/ObjWritingTests/Game/IW3/Menu/MenuDumperIW3Test.cpp
Laupetin: Seems to work fine, fine to merge then i suppose

Laupetin (approved):
Thank you very much for the implementation :)
Will also append a second PR for adjusting iw4 and iw5 menu code with improvements from this one in the coming days.

PR #917, opened 2026-07-18
fix: apply menu item origin as an offset

### LINKED ISSUE #78: [Menus] Custom menus of IW4x do not seem to result in the same
When trying to build IW4x's menus with OAT instead of building them with the game's menu compiler, some menus seem to be off.

### Serverlist

The serverlist cannot be used properly and the text on the "source" button is missing

image

### Options

The sidebar on the left seems to have a weird white glow.

image

michaeloliverx:
Add `origin` values to the existing item rectangle instead of replacing its coordinates.

Fixes https://github.com/Laupetin/OpenAssetTools/issues/78.

image

### DIFF
--- a/src/ObjLoading/Parsing/Menu/Sequence/ItemScopeSequences.cpp
+++ b/src/ObjLoading/Parsing/Menu/Sequence/ItemScopeSequences.cpp
@@ -255,8 +255,8 @@ namespace menu::item_scope_sequences
         {
             assert(state->m_current_item);
 
-            state->m_current_item->m_rect.x = MenuMatcherFactory::TokenNumericExpressionValue(state, result);
-            state->m_current_item->m_rect.y = MenuMatcherFactory::TokenNumericExpressionValue(state, result);
+            state->m_current_item->m_rect.x += MenuMatcherFactory::TokenNumericExpressionValue(state, result);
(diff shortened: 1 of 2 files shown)


Laupetin:
awesome :D yep that was definitely a mistake. The thing now is that what the game does is pretty weird, no? 🤔 
If you specify origin twice it will keep adding to the coordinates instead of replacing the origin.
But i guess minor detail and mistake of the game devs.

Also I am wondering how this interacts with `exp rect x`. When optimizing menus is enabled (default), it will inline any constant rect x into `item->window.rectClient.x`. That will override any value that was previously set by `origin`? But i guess that's the same for the vanilla behaviour?

Two things i'd usually not do but i guess the vanilla game does them so OAT should as well?

Laupetin:
Also now that i see it: Optimizing the `exp rect x` to be constant only writes in the `rectClient` and not `rect` as well? Also seems to be wrong to me 😬 But i guess that should maybe be fixed in another PR.

Laupetin (approved):
Thank you very much ❤️ Messed that up 😅 IW4x menus seem to be perfect to my eyes at least? Pretty cool!

Laupetin:
> Also now that i see it: Optimizing the `exp rect x` to be constant only writes in the `rectClient` and not `rect` as well? Also seems to be wrong to me 😬 But i guess that should maybe be fixed in another PR.

addressed that in #920

PR #918, opened 2026-07-19
feat: generate asset lists for linked zones

### LINKED ISSUE #21: Generate assetlists after linking with Linker
Assetlists are already usable for ignoring assets inside project definitions as well as loading all assets from an assetlist.

This is particularly useful when you want to build an IPak with all images from a built zone for example.

To enable this workflow Linker should automatically generate assetlists after successfully linking.
These assetlists should be placed in an appropriate source directory in a "assetlist" subfolder.

Would probably also good if creating assetlists could be turned off via cli args.
It should be enabled by default though for now.

### LINKED ISSUE #885: Feature Request: Assetlist output
It might be useful to some people to have the Linker output an asset list file of the built zone similar to how the original linker_pc did.

Example: `zone_source\english\assetlist\common_zombie.csv`
```
physpreset,bottle_plastic
physpreset,brick
etc
```

I am unsure of the use of this so I'm just putting the idea out there since it might have a use, this could also possibly be changed to work with IWDs and IPAKs also

michaeloliverx:
Fixes https://github.com/Laupetin/OpenAssetTools/issues/21 https://github.com/Laupetin/OpenAssetTools/issues/885

### DIFF
--- a/src/Linking/Linker.cpp
+++ b/src/Linking/Linker.cpp
@@ -8,6 +8,7 @@
 #include "SearchPath/SearchPaths.h"
 #include "Utils/Logging/Log.h"
 #include "Zone/AssetList/AssetList.h"
+#include "Zone/AssetList/AssetListOutputStream.h"
 #include "Zone/AssetList/AssetListReader.h"
 #include "Zone/Definition/ZoneDefinitionStream.h"
 #include "ZoneCreation/ZoneCreationContext.h"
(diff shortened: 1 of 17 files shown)


src/Linking/Linker.cpp
```diff
@@ -336,6 +337,38 @@ namespace
             return true;
         }
 
+        bool WriteAssetList(const Zone& zone) const
+        {
+            const auto sourceDirectory = fs::path(m_args.m_base_folder) / "zone_source";
```
Laupetin: I know that's part of the issue i wrote 3 years ago, but i think writing something in a source folder is not particularly cool 😅 Currently all output that is produced by OAT is in `zone_out` which is easy to gitignore or similar.
Writing stuff in other places as well complicates it a bit 😬 

So maybe `<outputfolder>/assetlist/<zone>.csv` would be an option?
What do you think?

michaeloliverx: Sounds good, I think that was were I intended to write it as well

PR #919, opened 2026-07-19
fix: handle UTF-8 BOM-prefixed input

### LINKED ISSUE #771: Parser does not correctly read files that are encoded with UTF-8 with BOM
Even though the usage of UTF-8 BOMs is not recommended by the unicode standard ("_Use of a BOM is neither required nor recommended for UTF-8_"), some software still love to create files that make use of it.
While OAT does not dump files with a BOM itself, it should be able to handle one if encountered.

See also: https://en.wikipedia.org/wiki/Byte_order_mark

Image
Image

michaeloliverx:
Fixes https://github.com/Laupetin/OpenAssetTools/issues/771

### DIFF
--- a/src/Parser/Parsing/IParserLineStream.cpp
+++ b/src/Parser/Parsing/IParserLineStream.cpp
@@ -1,5 +1,13 @@
 #include "IParserLineStream.h"
 
+#include <string_view>
+
+namespace
+{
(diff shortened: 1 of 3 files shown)


Laupetin (approved):
Works great :D Thank you very much. Not sure if this covers all files that are prone to this but i guess it already covers a good amount and it can be extended in the future

PR #922, opened 2026-07-20
chore: add Windows formatting convenience scripts

michaeloliverx:


Laupetin:
I always use the `.sh` scripts as I use zsh on Windows 😅 Batch scripts are cool for people that don't i guess but damn those are so much slower as they call clang-format for each file individually while the bash scripts call clang-format with all files at once.

On my PC that's the difference between 8 seconds and 36 seconds 😅 
I can't figure out how to make Windows append the files in one variable though 🤔

michaeloliverx:
Batch scripts are not the prettiest, how do you feel about powershell scripts instead? This one runs in about 6 seconds for me:
```ps1
# check-format.ps1

$ErrorActionPreference = "Stop"

$repositoryRoot = Split-Path -Parent $PSScriptRoot
$sourceDirectories = @(
    (Join-Path $repositoryRoot "src")
    (Join-Path $repositoryRoot "test")
)

$clangFormat = if ($env:CLANG_FORMAT_BIN) {
    $env:CLANG_FORMAT_BIN
}
else {
    "clang-format"
}

$files = @(
    Get-ChildItem -Path $sourceDirectories -Recurse -File |
        Where-Object { $_.Extension -in ".h", ".cpp" }
)

# Stay comfortably below Windows' native process command-line limit.
$batchSize = 100

for ($offset = 0; $offset -lt $files.Count; $offset += $batchSize) {
    $lastIndex = [Math]::Min($offset + $batchSize - 1, $files.Count - 1)
    $batch = $files[$offset..$lastIndex].FullName

    & $clangFormat -Werror -ferror-limit=1 --dry-run $batch

    if ($LASTEXITCODE -ne 0) {
        exit $LASTEXITCODE
    }
}
```

---

Also these could just be omitted altogether and a small guide added to the readme on running them through the provided scripts in a windows environment, or perhaps a dedicated `CONTRIBUTING.md` 👀

Laupetin:
> Batch scripts are not the prettiest, how do you feel about powershell scripts instead?

the reason i usually avoid ps1 scripts is because of Microsoft's ExecutionPolicy stuff always causing headaches and problems.
I don't know how it interacts with Git though, as I do not use powershell at all.
A quick test by creating the files and a commit, switching to another branch and back at least doesn't block execution for me (while having the default setting of `RemoteSigned`).
So I guess if it works via cloning with git, i'd have nothing against just using powershell scripts.

The script works well and definitely better than the batch version. I'd just add `#!powershell` as the first line so people like me can also execute it without having to start a new powershell session 😂

Laupetin:
> perhaps a dedicated CONTRIBUTING.md

i guess on the side of documentation, there's a lot of room for improvement 😅

Laupetin (approved):
works well from what i can tell, thanks :)

PR #923, opened 2026-07-21
feat: templated MapEnts dumper

michaeloliverx:

Removes more code than it adds :) I purposely didn't add a loader for now as the later games have additional binary data that cannot be handled at the moment, but at least we can dump all.


Laupetin (approved):
thanks :)

Issue #160, opened 2024-04-11
[Menus: IW5] One menu has a small issue


diamante0018:
I relinked every single menu on IW5 and although I might have missed a few I only found one error.

![Desktop Screenshot 2024 04 11 - 13 52 00 42](https://github.com/Laupetin/OpenAssetTools/assets/37080671/cd49c1f5-df17-426e-9742-90236b9b4e94)

The gameplay options menu is missing one background box for the game play options

The menu in question is settings_recipe_rules_base.menu (not 100% sure)

and can be found here https://github.com/diamante0018/iw5-menus/blob/main/zone_raw/localized_ui_mp/ui_mp/settings_recipe_rules_base.menu

diamante0018:
By the looks of it, it seems to be the same issue the server browser on iw4x (looks borked the same way)

michaeloliverx:
> Another instance of broken menus (for iw4): the cod4 menu remake: ![image](https://github.com/Laupetin/OpenAssetTools/assets/9197140/598c63be-83f0-4055-bd79-d15e859b8a90) https://github.com/reaaLx/iw3styledmenu
> 
> to be investigated. See also the [the discord discussion](https://discord.com/channels/1110531063161299074/1112314939902795907/1227942249292107860).

I hit several menu parser errors while linking [iw3styledmenu](https://github.com/reaaLx/iw3styledmenu/tree/master).

I used this command:

```bat
Linker --verbose --load "C:\Program Files (x86)\Steam\steamapps\common\Call of Duty Modern Warfare 2\zone\english\localized_ui_mp.ff" --menu-permissive IW3Menus
```

Some rejected forms come straight from stock [CoD4 menu source](https://github.com/promod/CoD4-Mod-Tools). Others follow normal preprocessor syntax. OAT rejects them or takes the wrong branch.

The stock examples below come from the CoD4 Mod Tools `raw` tree.

## 1. Report an unclosed block comment at its source

`ui_mp/menu_cac_assault.menu` includes `ui_mp/cac_loadout.inc`. A stray `/*` in that include caused this error:

```text
ERROR: In "ui_mp/menu_cac_assault.menu": Unclosed menu at end of file!
ERROR: Could not read menu file "ui_mp/menu_cac_assault.menu"
```

Example:

```c
itemDef
{
    text "before comment"
    /* comment with no end
}
```

OAT removed the rest of the file as comment text, reached EOF, and then blamed the menu braces.

OAT should report the comment start instead:

[...]

PR #924, opened 2026-07-21
fix(parser): report unclosed multiline comments

michaeloliverx:
When OAT builds this project:

https://github.com/reaaLx/iw3styledmenu/tree/master

It prints:

```
ERROR: In "ui_mp/menu_cac_assault.menu": Unclosed menu at end of file!
ERROR: Could not read menu file "ui_mp/menu_cac_assault.menu"
```

The menu has matching braces. The real fault lies in `ui_mp/cac_loadout.inc`, which ends with `/*` but has no `*/`. This makes OAT treat the rest of the menu, including its closing braces, as a comment.

This fix makes OAT track where each block comment starts. If OAT reaches EOF before `*/`, it now prints the right file and line:

```
ERROR: ui_mp/cac_loadout.inc L244:2 Unclosed multi-line comment
ERROR: Parsing menu file failed!
ERROR: Could not read menu file "ui_mp/menu_cac_assault.menu"
```

The new test covers an unclosed block comment at EOF.

Relates to https://github.com/Laupetin/OpenAssetTools/issues/160

### DIFF
--- a/src/Parser/Parsing/Impl/CommentRemovingStreamProxy.cpp
+++ b/src/Parser/Parsing/Impl/CommentRemovingStreamProxy.cpp
@@ -6,7 +6,7 @@ CommentRemovingStreamProxy::CommentRemovingStreamProxy(IParserLineStream* stream
     : m_stream(stream),
       m_inside_multi_line_comment(false),
       m_next_line_is_comment(false),
-      m_multi_line_comment_start_pos(0u)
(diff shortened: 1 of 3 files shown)


Laupetin (approved):
Thank you very much :) Just changed the variable names a bit since i got confused and hurt myself trying to merge the two variables that should not be merged

PR #926, opened 2026-07-21
fix(parser): parse spaced and negated defined checks

michaeloliverx:
Relates to No.3 in https://github.com/Laupetin/OpenAssetTools/issues/160#issuecomment-5035149309


Laupetin (approved):
Thank you very much :) The test is helping a lot understanding the code here

PR #927, opened 2026-07-21
fix(menu): allow numeric dvar values

michaeloliverx:
Relates to No.2 in https://github.com/Laupetin/OpenAssetTools/issues/160#issuecomment-5035149309


Laupetin (approved):
thanks 😄 i just replaced the usage of std::ostringstream and std::noshowpoint with std::format as that's generally what the project uses nowadays when there's a fixed amount of insertions in the string.

PR #929, opened 2026-07-22
feat: templated vehicle loading and dumping for IW4 and IW5

michaeloliverx:


test/ObjWritingTests/Game/IW4/Vehicle/VehicleDumperIW4Test.cpp
Laupetin, line 1: Generally the SystemTests are meant for tests that integrate more than one component 😅 
Since those are two separate tests that individually test loading and writing i'd put them in the respective component test project as unit tests.

src/ObjCommon/Game/IW4/Vehicle/VehicleStringsIW4.h
Laupetin, line 1: Added the strings into a separate file with also a game suffix so it's easier to identify them when they are tabs in the IDE.

Laupetin (approved):
thank you very much 😄 Split and expanded tests in an additional commit

PR #932, opened 2026-07-23
fix(menu): allow IW4 columns without alignment

michaeloliverx:
Relates to No.4 in https://github.com/Laupetin/OpenAssetTools/issues/160#issuecomment-5035149309


Laupetin (approved):
thanks :) I restructured the code a bit to also make it work for IW5.
And i made some local cleanups to move stuff into an anonymous namespace.

PR #933, opened 2026-07-23
fix(parser): allow space before function macro arguments

michaeloliverx:
Relates to No.5 in https://github.com/Laupetin/OpenAssetTools/issues/160#issuecomment-5035149309


Laupetin (approved):
didn't know this was valid before 😅 good find, thanks

PR #934, opened 2026-07-24
feat: T4 add `StringTable` loader

michaeloliverx:
The final game needed for full `StringTable` support.


Laupetin (approved):
thanks :) Looked through the implementations for the rest of the games as well and tried to make them match their behaviour

Issue #328, opened 2025-01-03
Add branding rawfile AssetCreator


Laupetin:
To be able to better recognize OpenAssetTools created files and help with support, a rawfile with the same name as the target should be added to the end of each fastfile.
The rawfile should include a branding text like "Created with OpenAssetTools vX.X.X" that includes the current OpenAssetTools version.
Similar to the branding text that is already added to ipaks.

This makes it possible to recognize OpenAssetTools created files in the wild and the respective version.
This helps with identifying mods that were created with an OpenAssetTools version that potentially has a certain bug.

An additional line of text should also be included to help identify the original source of the mod.
It should include a line like "NAME_OF_MOD by NAME_OF_AUTHOR"
* The name of the mod can be specified with the property  `>mod.name`
* The name of the author can be specified with the property  `>mod.author`
* As a fallback, if no name for the mod has been specified, it should take the name of the target
* There should be no fallback for the name of the author since there should be no identifying information unless explicitly wanted

diamante0018:
If I manually add rawfile,mod to my zone definition will it override this setting? What If I wanted to add epic XXX_gamertagXX custom rawfile with my own unique branding? 😢 

![image](https://github.com/user-attachments/assets/d61ece9f-c03f-4130-abd8-4af958bea84f)
![image](https://github.com/user-attachments/assets/935fa93a-0942-4e37-bc51-93ab4f2000cf)

Laupetin:
> If I manually add rawfile,mod to my zone definition will it override this setting? What If I wanted to add epic XXX_gamertagXX custom rawfile with my own unique branding? 😢

Then you can still create another rawfile that does not have this exact name 😅

diamante0018:
That is talk from someone who doesn't want to legalize nuclear bombs 💯

diamante0018:
Can you add meta tag >mod.description ? so I can add propaganda there?

michaeloliverx:
What do you think about using a more structured format for the rawfile contents? For example:
```
generator.name=OpenAssetTools
generator.version=v0.30.0
mod.name=My Mod
mod.author=Michael Oliver
```

This is inspired by the output of `git config --list`. The exact key names are only illustrative, the main idea is to use a simple, structured key/value format. It would also make the format easy to extend with optional metadata such as `mod.repo`, `mod.website`, or other useful fields in the future.

Laupetin:
[...]

PR #944, opened 2026-07-27
feat: add branding metadata rawfile to generated zones

michaeloliverx:
Relates to https://github.com/Laupetin/OpenAssetTools/issues/328

Example output:
```
generator.name=OpenAssetTools
generator.version=v0.31.0-28-gf99726d6
mod.name=iw4x_code_post_gfx_mp
```

Keys can be changed / discussed.

### DIFF
--- a/src/ObjCompiling/RawFile/BrandingAssetCreator.h
+++ b/src/ObjCompiling/RawFile/BrandingAssetCreator.h
@@ -18,13 +18,14 @@ namespace raw_file
 
     namespace detail
     {
-        template<typename RawFile> RawFile* CreateBrandingRawFile(MemoryManager& memory, const std::string& name, const std::string& brandingText)
+        template<typename RawFile, bool Compress>
+        RawFile* CreateBrandingRawFile(MemoryManager& memory, const std::string& name, const std::string& brandingText)
         {
             auto* rawFile = memory.Alloc<RawFile>();
             rawFile->name = memory.Dup(name.c_str());
             rawFile->len = static_cast<int>(brandingText.size());
 
-            if constexpr (requires { rawFile->compressedLen; })
+            if constexpr (Compress)
             {
                 auto compressedSize = compressBound(static_cast<uLong>(brandingText.size()));
                 auto* compressedBuffer = memory.Alloc<char>(compressedSize);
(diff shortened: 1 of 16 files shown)


src/ObjCompiling/RawFile/BrandingAssetCreator.h
```diff
@@ -0,0 +1,98 @@
...
+    {
+        template<typename RawFile> RawFile* CreateBrandingRawFile(MemoryManager& memory, const std::string& name, const std::string& brandingText)
+        {
+            auto* rawFile = memory.Alloc<RawFile>();
+            rawFile->name = memory.Dup(name.c_str());
+            rawFile->len = static_cast<int>(brandingText.size());
+
+            if constexpr (requires { rawFile->compressedLen; })
+            {
+                auto compressedSize = compressBound(static_cast<uLong>(brandingText.size()));
```
michaeloliverx, line 30: Maybe we should just not compress this asset?

Laupetin: yeah no point in compressing it 😅 more work to read the data and it wouldn't save a lot of space anyway

src/ObjCompiling/RawFile/BrandingAssetCreator.h
```diff
@@ -0,0 +1,101 @@
...
+            else
+            {
+                auto* buffer = memory.Alloc<char>(brandingText.size() + 1u);
+                std::memcpy(buffer, brandingText.data(), brandingText.size());
+                buffer[brandingText.size()] = '\0';
+
+                if constexpr (requires { rawFile->compressedLen; })
+                    rawFile->compressedLen = 0;
+
+                if constexpr (requires { rawFile->data.buffer; })
```
Laupetin, line 53: TIL: you can do this... Oh wonders of modern c++.

Still a bit icky in bigger components over multiple games because its hard to debug which configuration gets what result 😅 but it's okay here i guess

src/ObjCompiling/RawFile/BrandingAssetCreator.h
```diff
@@ -0,0 +1,101 @@
...
+        AssetCreationResult CreateAsset(const std::string&, AssetCreationContext&) override
+        {
+            return AssetCreationResult::NoAction();
+        }
+
+        void FinalizeZone(AssetCreationContext& context) override
+        {
+            const auto brandingText = CreateBrandingText(m_zone_definition, m_zone.m_name);
+            auto* rawFile = detail::CreateBrandingRawFile<typename Asset_t::Type, Compress>(m_memory, m_zone.m_name, brandingText);
+            context.AddAsset<Asset_t>(m_zone.m_name, rawFile);
```
Laupetin, line 87: This should probably check if such a file is already in the zone?
This current overwrites any rawfile with the same name.

It would probably be better to prioritize user input and assets over this branding file.

PR #945, opened 2026-07-28
feat(unlinker): support `?game?` placeholder in output folder

michaeloliverx:
Should we also add `?platform?`?


test/UnlinkingTests/UnlinkerArgsTests.cpp
```diff
@@ -0,0 +1,45 @@
+#include "UnlinkerArgs.h"
```
michaeloliverx, line 1: Not sure where these tests should live.

test/UnlinkingTests/UnlinkerArgsTests.cpp
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    TEST_CASE("Unlinker output folder replaces game and zone placeholders", "[unlinker][arguments]")
+    {
+        UnlinkerArgs args;
+        args.m_output_folder = "zone_dump/?game?/?zone?";
+
+        const Zone zone("test_zone", 0, GameId::IW4, GamePlatform::PC);
+
+        REQUIRE(args.GetOutputFolderPathForZone(zone) == "zone_dump/iw4/test_zone");
```
michaeloliverx, line 43: Personally I wouldn't mind this being the default behaviour.

test/UnlinkingTests/UnlinkerArgsTests.cpp
Laupetin: i created a new UnlinkingTests component for it since that's what i did previously. Maybe a bit too many test projects for very little tests but that's the architecture right now i guess 😅

test/UnlinkingTests/UnlinkerArgsTests.cpp
Laupetin: yes me neither. I am afraid of breaking scripts people made though 😅 So im a bit torn on that one

Laupetin:
> Should we also add `?platform?`?

In the future when there are multiple platforms: sure

For now: I am unsure whether "platform" is the word that should be used?
It makes sense when talking about xenon <> pc but what about cod:ol and all of its different versions?
And iw4x PC and MS Store being x86 and x64?

Is "platform" the right term then? 😅

test/UnlinkingTests/UnlinkerArgsTests.cpp
michaeloliverx: Technically pre 1.0 allows for such changes 😛

Laupetin: hm okay, but will do separately

Laupetin:
thanks btw :) 😅

PR #947, opened 2026-07-31
fix(menu): accept numeric `dvarStrList` values

michaeloliverx:
Convert numeric tokens to strings to match the game menu parser behaviour.

For example, this is accepted by the IW3/IW4 game parser and the IW3 linker:
```
dvarStrList {
    "@MPUI_RULES_UNLIMITED" 0
    "@MPUI_RULES_5MINUTES" 5
    "@MPUI_RULES_10MINUTES" 10
}
```

Numeric values such as `5` are stored internally as strings, equivalent to `"5"`.

https://github.com/Laupetin/OpenAssetTools/issues/160

Thank you @reaaLx for reporting.

### DIFF
--- a/src/ObjLoading/Parsing/Menu/Matcher/MenuMatcherFactory.cpp
+++ b/src/ObjLoading/Parsing/Menu/Matcher/MenuMatcherFactory.cpp
@@ -67,7 +67,7 @@ MatcherFactoryWrapper<SimpleParserValue> MenuMatcherFactory::TextNoChain() const
     }));
 }
 
-MatcherFactoryWrapper<SimpleParserValue> MenuMatcherFactory::TextOrNumericNoChain() const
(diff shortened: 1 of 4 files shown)


src/ObjLoading/Parsing/Menu/Matcher/MenuMatcherFactory.cpp
```diff
@@ -67,6 +67,22 @@ MatcherFactoryWrapper<SimpleParserValue> MenuMatcherFactory::TextNoChain() const
     }));
 }
 
+MatcherFactoryWrapper<SimpleParserValue> MenuMatcherFactory::TextOrNumericNoChain() const
```
Laupetin: I think a better name would be `TextNoChainOrNumeric`?
The "no chain" part is only important for the "text" part.

It makes it so `"foo" " " "bar"` becomes `"foo bar"`. Like it does in C.
And "no chain" means it should not do that.

Laupetin (approved):
Thanks :)

PR #948, opened 2026-07-31
fix(parser): allow empty final macro arguments

michaeloliverx:
Allow function-like macros to have an empty final argument, matching the game menu preprocessor.

IW4 menu macros can expand to calls such as:

```text
CHOICE_SCRIPTS_ALL(..., ; CHOICE_GETFOCUSEDITEMY, )
```

OAT previously rejected the closing `)` after the comma:

```text
Cannot close macro parameters after comma
```

The final argument is now treated as empty.

Thank you @reaaLx for the report!

https://github.com/Laupetin/OpenAssetTools/issues/160


Laupetin (approved):
Makes sense, whoops 😅 

Thanks ❤️

PR #966, opened 2026-08-10
build: support macOS x86_64 and arm64

michaeloliverx:
I am not sure if this is something you are interested in, but this at least can serve as a record for what changes were needed to successfully compile on macOS :)

Both the x86_64 (older Intel based systems) and the arm64 builds compile successfully and run on my system. Apple is ending support for Intel x86 architecture, with upcoming macOS releases dropping hardware compatibility for Intel-based Macs entirely so that branch might not even be needed.


Laupetin:
The big issue i have with this is the following:
- Apple is the only vendor that locks down their OS and devices so much that it is not possible to either emulate or virtualize any machine that could serve as a testing field for compiling the code and testing it or fixing it in case something breaks.
- I do not have any Apple device

That means I cannot at all maintain any code that is specific to apple and i cannot at all fix issues that arise from Apple standard library or compiler not liking a change.
And i am not really comfortable with having code that i cannot touch or edit at all 😬 

With Linux i can either spin up WSL or use a device of mine that runs it.
Which is what i usually just do when Linux compilation decides to fail when I push some changes i did when on Windows,

PR #976, opened 2026-08-21
fix: handle IWD entry end of stream

michaeloliverx:
Fixes the IWD stream wrapper’s EOF and peek handling.

Minizip returns `0` at the end of an archive entry, but the wrapper treated it as a successful read and could replay a stale byte. Bulk reads after `peek()` also failed to account for the cached byte correctly.

This path is not currently exercised by OpenAssetTools atm, but the issue was observed in a fork while consuming an IWD entry to EOF. This is a general correctness fix for any IWD-backed stream consumer.

Added an IWD regression test covering peek → bulk read → EOF.

### DIFF
--- a/test/ObjCommonTests/SearchPath/IwdTests.cpp
+++ b/test/ObjCommonTests/SearchPath/IwdTests.cpp
@@ -1,0 +1,132 @@
+#include "OatTestPaths.h"
+#include "SearchPath/IWD.h"
+
+#include <array>
+#include <catch2/catch_test_macros.hpp>
+#include <cstdint>
+#include <filesystem>
+#include <fstream>
+#include <string_view>
+
+namespace fs = std::filesystem;
+
+namespace
+{
+    void WriteUint16(std::ostream& stream, const std::uint16_t value)
+    {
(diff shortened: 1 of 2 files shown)


Laupetin (dismissed):
good find, thank you very much :)

test/ObjCommonTests/SearchPath/IwdTests.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+        // End of central directory.
+        WriteUint32(stream, 0x06054B50u);
+        WriteUint16(stream, 0u);
+        WriteUint16(stream, 0u);
+        WriteUint16(stream, 1u);
+        WriteUint16(stream, 1u);
+        WriteUint32(stream, centralDirectorySize);
+        WriteUint32(stream, centralDirectoryOffset);
+        WriteUint16(stream, 0u);
```
Laupetin: i got rid of this though and just used minizip to write the zip file 😅 
Because why not use the dependency when you already have it.

And this code is a bit hard to verify whether it works correctly

PR #980, opened 2026-08-22
feat: decode IWI6 wavelet textures

michaeloliverx:
Needs #976.

The wavelet decoder consumes IWD-backed image streams through EOF, so #976 is required to prevent the IWD stream wrapper from repeatedly returning a stale byte at the end of an entry.

Examples you can check (IW3):

| Fastfile       | Wavelet image    | Format        |
| -------------- | ---------------- | ------------- |
| `common.ff`    | `spotlight_beam` | 9 — luminance |
| `common_mp.ff` | `spotlight_beam` | 9 — luminance |
| `ui.ff`        | `logo_cod2`      | 6 — RGBA      |
| `ui_mp.ff`     | `logo_cod2`      | 6 — RGBA      |

AI assistance was used to help translate and validate the Huffman table representation. Rather than copying the three 4,096-entry lookup tables verbatim, this implementation stores their compact codeword definitions and generates the lookup tables at compile time. The generated tables were compared entry-for-entry against the reference tables, with all 12,288 entries matching exactly.

KisakCOD helped a lot here. See https://github.com/SwagSoftware/KisakCOD/blob/1c03702cbe176e9274e486d295edcd035b3c2b5f/src/gfx_d3d/r_imagedecode.cpp#L395-L414

### DIFF
--- a/src/ObjImage/Image/IwiWaveletDecoder.cpp
+++ b/src/ObjImage/Image/IwiWaveletDecoder.cpp
@@ -1,0 +1,458 @@
+#include "IwiWaveletDecoder.h"
+
+#include "Image/ImageFormat.h"
+#include "Utils/Logging/Log.h"
+
+#include <algorithm>
+#include <array>
+#include <bit>
+#include <cstdint>
+#include <iterator>
+#include <limits>
+#include <vector>
+
+namespace
+{
+    constexpr auto HUFFMAN_LOOKUP_BITS = 12u;
+    constexpr auto HUFFMAN_LOOKUP_SIZE = 1u << HUFFMAN_LOOKUP_BITS;
+    constexpr auto ESCAPE_VALUE = std::numeric_limits<std::int16_t>::min();
+
+    struct HuffmanCodeword
+    {
+        std::uint16_t code;
+        std::uint8_t bit_count;
+        std::int16_t value;
+    };
+
+    struct HuffmanDecode
+    {
+        std::int16_t value;
+        std::uint8_t bit_count;
+    };
+
+    // These are the unique, LSB-first codewords from the IW wavelet decoder's
+    // three 4096-entry lookup tables. The full lookup tables are generated at
+    // compile time below.
+    constexpr std::array<HuffmanCodeword, 180> BLUE_CODEWORDS{
+        {
+         {0x001, 3, 0},    {0x004, 5, 4},    {0x005, 5, 2},    {0x007, 5, 1},    {0x00A, 5, 3},    {0x014, 5, -4},
+         {0x015, 5, -2},   {0x017, 5, -1},   {0x01A, 5, -3},   {0x000, 6, 12},   {0x002, 6, 10},   {0x003, 6, 7},
+         {0x006, 6, 9},    {0x00B, 6, 6},    {0x018, 6, 11},   {0x01E, 6, 8},    {0x01F, 6, 5},    {0x020, 6, -12},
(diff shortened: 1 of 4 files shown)


src/ObjImage/Image/IwiWaveletDecoder.cpp
```diff
@@ -0,0 +1,479 @@
...
+            if (entry.bit_count == 0)
+                throw std::runtime_error("Incomplete Huffman lookup");
+        }
+
+        return result;
+    }
+
+    constexpr auto BLUE_LOOKUP = CreateHuffmanLookup(BLUE_CODEWORDS);
+    constexpr auto RED_GREEN_LOOKUP = CreateHuffmanLookup(RED_GREEN_CODEWORDS);
+    constexpr auto ALPHA_LOOKUP = CreateHuffmanLookup(ALPHA_CODEWORDS);
```
Laupetin, line 147: fancy and smart, not bad.

Did you create a little script to generate these codeword tables?
Or how did you calculate them?

Maybe as a reference for future iwi version support 😅 
Or do all iwi version use the same huffman tables? 🤔

src/ObjImage/Image/IwiWaveletDecoder.cpp
```diff
@@ -0,0 +1,479 @@
...
+                return false;
+        }
+
+        return true;
+    }
+
+    constexpr int ShiftRightOne(const int value)
+    {
+        return value >= 0 ? value / 2 : -((-value + 1) / 2);
+    }
```
Laupetin: Why that? The game seems to do `>> 1` here?
Also i know there is a difference between arithmetic and logical shift but this does neither?

Image

What is the reason for this?

michaeloliverx: Ah yeah, the intention was just to preserve the same behaviour for negative values. It does end up equivalent to the >> 1 here?, but looking at it again it’s definitely an unnecessarily complicated way of doing it 😅

wroyca:
For reference, here's the implementation for IW4:

https://github.com/iw4x-x64/oat/blob/544386887aabce5bf3c242df3896821cc39f61c1/src/ObjImage/Image/IwiWaveletDecoder.cpp
https://github.com/iw4x-x64/oat/blob/544386887aabce5bf3c242df3896821cc39f61c1/src/ObjImage/Image/IwiWaveletCodebooks.h

src/ObjImage/Image/IwiWaveletDecoder.cpp
Laupetin: seems like all relevant iwi versions do use these huffman tables (iwi6, iwi8, iwi13). iwi27 with T6 does not support wavelet formats at all on PC, so whether it would usually use the same tables is unknown.

src/ObjImage/Image/IwiWaveletDecoder.cpp
Laupetin: It is almost identical to `>> 1` here. As seen in the screenshot, it handles the edge cases differently of e.g. INT_MIN.
I changed it to `>> 1` now to match the game.

Laupetin:
> For reference, here's the implementation for IW4:
> 
> https://github.com/iw4x-x64/oat/blob/544386887aabce5bf3c242df3896821cc39f61c1/src/ObjImage/Image/IwiWaveletDecoder.cpp https://github.com/iw4x-x64/oat/blob/544386887aabce5bf3c242df3896821cc39f61c1/src/ObjImage/Image/IwiWaveletCodebooks.h

Thanks :) You told me but i honestly forgot you said you also had an implementation 😅 The one from this PR works fine for IW3, IW4, T4 and T5 but having yours as a resource in the PR might be nice for future reference

Laupetin (approved):
Thank you very much <3 No more unsupported iwi format errors 🎉

PR #989, opened 2026-09-04
feat: T4 add xmodel loader

michaeloliverx:
Linked a test mod.ff and replaced the thompson with the legendary banana seen in https://github.com/OpenAssetTools/Examples.

image


Cowldwink:
The GOAT gun made another peak cameo in another title. Cooool 😎

src/ObjLoading/XModel/LoaderXModel.cpp.template
Laupetin: This is also the case for T4, this is writing random memory to the fastfile 😅

Laupetin: Same thing here

Laupetin (approved):
🍌

PR #990, opened 2026-09-04
fix: correct T4 FX element type values

michaeloliverx:
T4's `linker_pc.exe` has this:
```c++
  v2 = *(_BYTE *)(a2 + 452);
  ...
  if ( v2 == 4 )
  {
    v4 = *(_DWORD *)(a2 + 2140);
    if ( !v4 || (v5 = *(_DWORD *)(a2 + 1880)) == 0 )
    {
      Com_PrintError(21, "effect '%s' segment '%s'\nTrail cross-section cannot be empty.\n", a1, (const char *)a2);
      return 0;
    }
```

I think this indicates that type `4` is a trail?


Laupetin (dismissed):
yes, very correct. T4 also seems to contain code to load obj fx files, that code confirms the exact enum you just made.
Good catch, thanks :)

PR #1004, opened 2026-09-13
fix(techset): write complete Direct3D 9 shader bytecode

LVEIceFire:
## Summary

- use the shader program element size when writing Direct3D 9 bytecode
- avoid truncating shader output to the host pointer size
- add a regression test that verifies the complete byte sequence

## Validation

- targeted ObjWriting test: 78 assertions in 1 test case
- `git diff --check`

The test data is synthetic and contains no game assets.

## Related independent PRs

- #1006 — fix(zone): align loaded block buffers
- #1007 — feat(iw5): support updated PC x64 fastfiles


Laupetin (approved):
Good catch, that's a hard to spot typo. Even though it takes the size of the wrong thing, the values match in x86, so it only manifests in x64, where it is now fixed.

Thank you for contributing a fix!

Laupetin:
I have to say though: At the very least this PR description looks very AI generated.
It includes unnecessary and out of place statistics, states some very odd "validation" and links to PRs that are not related?

I have now written contribution guidelines, they did not exist at the creation of this PR though. But please consider those in the future.

PR #1006, opened 2026-09-13
fix(zone): align loaded block buffers

LVEIceFire:
## Summary

- allocate loaded zone block buffers with explicit 16-byte alignment
- retain matching aligned deallocation through an owning smart pointer
- cover small, boundary-sized, large, and empty allocations

## Motivation

During the MW3 x86 corpus run, 13 legacy material exports failed at aligned `MaterialConstantDef` accesses even though the stream offsets were aligned. The common factor was the base address of `XBlock::m_buffer`. This change makes the required 16-byte base alignment explicit and pairs the aligned allocation with matching deallocation.

## Validation

- targeted SystemTests test: 10 assertions in 1 test case
- `git diff --check`

## Related independent PRs

- #1004 — fix(techset): write complete Direct3D 9 shader bytecode
- #1007 — feat(iw5): support updated PC x64 fastfiles


test/ZoneCommonTests/Zone/XBlockTests.cpp
Laupetin, line 1: This is not in the right test project. It should be in ZoneCommonTests.

src/ZoneCommon/Zone/XBlock.cpp
```diff
@@ -1,5 +1,17 @@
 #include "XBlock.h"
 
+#include <new>
+
+namespace
+{
+    constexpr std::align_val_t XBLOCK_BUFFER_ALIGNMENT{16u};
```
Laupetin: While this is in fact the highest alignment that any struct in the game headers do, i've checked on what the actual game code does and it always seems to align with 4096 (0x1000) so i'd suggest to just mirror the behaviour to avoid having some situation in the future where this might cause trouble again and to have parity with the game.

Laupetin (dismissed):
Damn, okay, i have never noticed or thought about it 🤔 But yeah i guess just aligning the block offset is of no use if the base address is not aligned. And aligning by address would also be wrong as that could create unexpected holes in the block and potentially overflow it.

Very interesting, thanks for providing a fix!

Laupetin:
https://github.com/Laupetin/OpenAssetTools/pull/1004#issuecomment-5752831651 also applies here

PR #1007, opened 2026-09-13
feat(iw5): support updated PC x64 fastfiles

LVEIceFire:
## Summary

- describe the mixed-width serialized ABI used by updated IW5 PC x64 fastfiles
- generate loading code for native pointers with 32-bit serialized offsets and aliases
- update affected IW5 asset layouts and reusable pointer-array handling
- write updated IW5 PC x64 fastfiles with matching pointer-slot and offset widths
- add focused ABI, alias, and zone round-trip regression coverage

## Validation

- standalone SystemTests: 254 assertions in 19 test cases
- IW5 simple-zone round trip: 9 assertions in 1 test case
- listed the official German `sp_ny_manhattan.ff` with 0 warnings and 0 errors
- `git diff --check`

No proprietary game files are included.

## Related independent PRs

- #1004 — fix(techset): write complete Direct3D 9 shader bytecode
- #1006 — fix(zone): align loaded block buffers


Kenshin9977:
Hi, I tested this PR and it works. Two small things:

1. With Visual Studio 2022 the build fails on warning C4267 in the generated loader code (a `size_t` passed as an `unsigned int`), since warnings are treated as errors. A cast would fix it.
2. In the x64 sound info, the first bytes are a standard WAVEFORMATEX header (16-bit format, 16-bit channels, rate, bytes per second, 16-bit block align, 16-bit bits). The current struct reads format and channels as one int, so dumped sounds get the block align as their channel count and the byte rate as their sample rate (a mono 44100 Hz sound comes out as stereo 88200 Hz). Changing the start of the struct to match fixes it.

LVEIceFire:
Thanks for checking this and for pointing out the two issues. I’ve pushed fixes for both:

- the generated loader now casts the pointer-byte count passed to the unsigned parameter, which addresses C4267 in Visual Studio 2022;
- the x64 `AILSOUNDINFO` prefix now follows the 16-byte PCM/WAVE field layout. I also added checks for that layout and for exporting a mono 44.1 kHz WAV.

While rerunning the Visual Studio checks, I found an x64 alignment mismatch in two generated XAnim structure tests and corrected that too.

In this validation run, the Visual Studio 2022 x86 and x64 jobs completed with warnings treated as errors, including the repository’s test step:
https://github.com/LVEIceFire/OpenAssetTools/actions/runs/35092723453

I also repeated the `sp_ny_manhattan` export. It produced 143 WAV files, and the header and size checks matched for the files in that run.

Could you try the original build and export again when you have a chance?

Kenshin9977:
Thanks, the build is clean now on VS2022 x64 with warnings as errors, and the sound fix looks right on my side.
One problem with the last commit (native x64 xanim alignment): it breaks MW3 loading completely. Every zone segfaults, even with --list. The commit before it works fine.
Removing type_align32(4) at x64 does not only change the allocation alignment, it also changes the generated fill reads: AppendToFill(4) becomes AppendToFill(8), and AtOffset(4) becomes AtOffset(8) for XAnimPartTrans and XAnimDeltaPartQuat2, so the union is read 4 bytes too far. Adding set allocalign for those types does not compensate, since that only affects the Alloc call. The struct sizes are the same with and without the attribute at x64, so the header change only moves the generator. Fixing the struct test expectation instead may be the way

Laupetin:
hey there, thanks for taking the time to come up with a solution to how IW5 x64 could be supported.
The thing is that this implementation does not really align with how I want to generally approach multi platform/architecture support in OAT in the future which is to handle them as multiple variants of the same game most likely.

By making the x86 and x64 support different game version however, you are breaking compatibility of the x64 build with the x86 game version which it previously had.
This is not only a problem because the goal is to at some point switch the main OAT release build from x86 to x64 (this solution would require to offer both and have both support different games).
But it also will never allow for any x86 and x64 conversion which is already being used with IW4.

Again, cool that you got something working and thanks for making the effort to upstream it, but unfortunately for the reasons above I won't be able to use this solution for the upstream repo.
Since it's also a bigger topic concerning console support and game variations like iw4 and cod:ol as well, i currently looking into a solution myself that can tackle this on a broader scope.

LVEIceFire:
I’d like to take another shot at this. Are you already working on the variant side, or would a draft from me be useful?

PR #1008, opened 2026-09-15
feat(t5-t6/fx): add .efx effect file dumper and loader/linker

Gustavo10Destroyer:
Add FxEffectDumperT6 to export .efx text files from compiled T6
FxEffectDef assets. Registers in ObjWriterT6 asset dumper collection.

Based on old Greyhound implementation.
It also supports T5.


Gustavo10Destroyer:
Please forgive me for the PR—I had asked the AI to do it, but it was extremely sloppy work. I’m still working on this feature, but now I’m reviewing every step and testing everything before assuming it’s done.

The effects are quite complex, and using old Greyhound code may not have been a good choice because now that I've stopped to ask myself, “Why was support removed if this was ‘already done’?” I've figured out why.

PR #1009, opened 2026-09-16
fix: handle non-contiguous xmodel root bones

michaeloliverx:
Fixes GLTF/GLB XModel dumping when a malformed parent offset produces a non-contiguous root bone.

The exporter now:

- Treats invalid negative parent indices as root bones.
- Handles multiple, non-contiguous root bones correctly.
- Omits the optional GLTF skin skeleton field when there is no single root.
- Adds a regression test covering the affected T4 model.

The issue was observed when unlinking the mod.ff from this T4 map: https://callofdutyrepo.com/2020/04/15/spongebob-battle-for-bikini-bottom/


Laupetin:
I wonder how this even happens 🤔 exactly two models have these bad parent bones. I wonder if that's a bug in the xmodel export loader code of the mod tools then since `-1` would usually indicate a root bone.
Otherwise not sure what would even cause this.

Also nice `mod.exe` on the download 😂 Love to see it

Laupetin (approved):
thanks a lot :)

michaeloliverx:
> Also nice mod.exe on the download 😂 Love to see it

I was super sus on this as well, but apparently it's just how CoD5 custom zombies is packaged? I think they all use the same tool to package because the exe installer is the same style for all.

Laupetin:
> > Also nice mod.exe on the download 😂 Love to see it
> 
> I was super sus on this as well, but apparently it's just how CoD5 custom zombies is packaged? I think they all use some tool to package because the exe installer is the same style for all.

"Rightclick -> 7-Zip -> Open" to the rescue 😛 idk who came up with the idea of making self-extracting archive exes, but i'm not running those 😂

PR #1012, opened 2026-09-17
feat: templated FX loading and dumping IW3,T4,T5

michaeloliverx:
@SadSlothXL asked me to take a look into this for T5/T6. I only focused on the games that have official mod tools so I could check their exact behaviour.

`FxImpactTable` loading is intentionally left for a follow-up. The stock linker constructs the table by merging all CSV files in `fx/` and then `fx/maps/<name>/` in filename order, whereas the dumper emits one flattened resolved table. Supporting this cleanly requires deciding whether OAT should reproduce the stock layered-source behaviour, load only the flattened representation, or support both.


Cowldwink:
Seems to load T6 `fx`as intended on T5!
image
image

And oh boy this error saved me quite some time pin pointing the issue earlier.
image

However, it only errors on if your material's atlas *falls short* and not if the material atlas is *higher* than what the `fx` uses but it's okay 👍

